PR #16, opened 2025-12-15
Building on Nix/NixOS.

copeison:
# Summary

Basic support for building on Nix.

Due to limited experience with flakes i cannot add support for it.
Someone else would have to.

# Testing

Tested on latest NixOS unstable.

Everything has been tested to work besides 2 minor issues.

1. Auto device selection.
2. Alsa throwing a warning. (Has no effect on usability/playback.)

Besides that everything works!

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,11 @@ sudo dnf install openssl-devel alsa-lib-devel pkg-config
     cargo run --release
     ```
 
+    **Nix:**
+    ```bash
+    nix-build
+    ```
+
 ### Updating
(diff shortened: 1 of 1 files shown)


LargeModGames:
Is the version hardcoded? Like does it need to be manually updated everytime there is a new release?

copeison:
Yes it's typically hard coded. This is to make sure that builds will always have proper reproducibility. It's also hard coded to make sure that the exact version can be pulled.

Issue #80, opened 2026-02-06
Slim build (--no-default-features --features telemetry) fails to compile


bodoque-01:
**Describe the bug**

Building with the slim build flags recommended in CONTRIBUTING.md fails with 2 compilation errors and 3 warnings.

The command from CONTRIBUTING.md:

```bash
cargo clippy --no-default-features --features telemetry
```

produces:

- `error[E0609]`: `auto_select_streaming_device()` in `src/network.rs` references `self.streaming_player`, which doesn't exist on `Network` without the streaming feature
- `error[E0282]`: type annotation failure (consequence of the above)
- `warning`: unused imports `Duration`, `Instant`, `Ordering` in `src/main.rs` (only used in `#[cfg(feature = "streaming")]` blocks)
- `warning`: unused variable `shared_position` in `src/main.rs` (only used with streaming)
- `warning`: unnecessary `mut` on variable `c` in `src/network.rs` (only mutated with streaming)

**To Reproduce**

1. Clone the repo at commit `750c5d3`
2. Run:
   ```bash
   cargo clippy --no-default-features --features telemetry
   ```
3. Compilation fails

**Expected behavior**

The slim build should compile successfully, as documented in CONTRIBUTING.md

**Screenshots**

**Desktop (please complete the following information):**

- OS: Linux (Ubuntu, kernel 6.8.0-90)
- Rust: stable 1.88.0
- Version: 0.35.5

**Additional context**

The CI workflow (`ci.yml`) only runs clippy with default features:

```bash
cargo clippy --locked -- -D warnings
```

This is why this wasn't caught. The slim build clippy command in CONTRIBUTING.md is not tested in CI.

Besides the bigger issue (failing to compile), running the slim-build clippy after fixing these errors reveals ~100 pre-existing `uninlined_format_args` warnings across the codebase, which fail with `-D warnings` (due to being promoted to errors).

TODO: I will open the PR in a few minutes, it addresses the harder errors that don't allow the app to compile.

bodoque-01:
Here is the PR: https://github.com/LargeModGames/spotatui/pull/81 the fix is very small, just adding "#[cfg(feature = "streaming")]" when it's applicable

PR #81, opened 2026-02-06
fix: solve slim-build-compilation error

bodoque-01:
# Summary

Fixes https://github.com/LargeModGames/spotatui/issues/80, the slim build (`--no-default-features --features telemetry`) from CONTRIBUTING.md fails to compile.

The root cause is that `auto_select_streaming_device()` and some imports/variables are only meaningful with the `streaming` feature but weren't properly feature-gated.

Changes:

- Gate `auto_select_streaming_device()` and its match arm behind `#[cfg(feature = "streaming")]`, with a no-op fallback for non-streaming builds

#

The following three changes are to just fix compilation warnings that appeared alongside the error.

- Move `Duration`, `Instant`, `Ordering` imports behind `#[cfg(feature = "streaming")]` in `main.rs`
- Prefix unused `shared_position` parameter with `_` in non-MPRIS `start_ui`
- Add `#[allow(unused_mut)]` on match arm where `c` is only mutated in streaming-gated code

# Testing

```sh
cargo check --no-default-features --features telemetry (compiles)
cargo fmt --all -- --check
cargo test --no-default-features --features telemetry     (49 passed, 0 failed)
```

# Additional notes

No logic or behavior changes, this is purely conditional compilation hygiene. The default (full-feature) build is unaffected.

Note: after fixing these compilation errors, running `cargo clippy --no-default-features --features telemetry -- -D warnings` reveals ~100 pre-existing `uninlined_format_args` warnings across the codebase (also present on the default-feature build). This is a separate issue from the compilation failures fixed here.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -75,11 +75,13 @@ use std::{
   io::{self, stdout, Write},
   panic,
   path::PathBuf,
-  sync::{
-    atomic::{AtomicU64, Ordering},
-    Arc,
-  },
-  time::{Duration, Instant, SystemTime},
+  sync::{atomic::AtomicU64, Arc},
+  time::SystemTime,
+};
+#[cfg(feature = "streaming")]
+use std::{
+  sync::atomic::Ordering,
+  time::{Duration, Instant},
 };
 use tokio::sync::Mutex;
 use user_config::{UserConfig, UserConfigPaths};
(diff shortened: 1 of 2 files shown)


LargeModGames:
@all-contributors please add @bodoque-01  for code

PR #89, opened 2026-02-09
fix: handle help menu as a route to avoid closing the app on dismiss

dpnova:
# Summary

The second time you close the help screen it closes the whole app. 

This moves HelpMenu to be a route in its own right to avoid accidentally popping off an empty stack when you hit "q" on the second opening of help.

Also added some other tests for HelpMenu

# Testing

- cargo fmt --all
- cargo test --locked
test result: ok. 51 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s


LargeModGames:
@all-contributors please add @dpnova for code

PR #94, opened 2026-02-12
fix: Search input horizontal scroll

bodoque-01:
# Summary

Fixes the search input box (`/`) losing the cursor when text overflows the visible width. Before, typing a long query would cause the text to be clipped by the `Paragraph` widget while the terminal cursor kept advancing past the box boundary, making it impossible to see what you're typing actively typing beyond the width.
Now, the cursor simply follows the text being typed.

Changes:

- Add `input_scroll_offset: Cell<u16>` to `App` to track horizontal scroll state across render and cursor-positioning phases
- Compute a scroll offset in `draw_input_and_help_box` when `input_cursor_position` exceeds the inner width of the input box, and apply `.scroll((0, offset))` to the `Paragraph` widget so text pans left
- Subtract the scroll offset from the terminal cursor X coordinate in both `start_ui` cursor-positioning sites so the blinking cursor stays aligned with the visible text

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -22,6 +22,7 @@ use rspotify::{
   },
   prelude::*, // Adds Id trait for .id() method
 };
+use std::cell::Cell;
 use std::sync::mpsc::Sender;
 #[cfg(feature = "streaming")]
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


bodoque-01:
Ah I'll fix those and open a new PR

PR #97, opened 2026-02-13
Fixed Windows compilation bug introduced in v0.36.0

tylerdotrar:
# Summary

Version **v0.36.0** introduced a minor issue that resulted in the project not successfully building on Windows hosts.  A Linux-only feature -- `mpris` -- was added to the default list of features despite being internally labeled as optional.

So specifically, all this PR does is remove the optional Linux-only feature `mpris` from the default list within `Cargo.toml`, fixing Windows compilation.

# Testing

Compilation failure at line 302 of `main.rs` due to being unable to locate `mpris` on a Windows host.
```
cargo install spotatui
```
image

---

Root cause being this "optional" feature being included in the "default" list within `Cargo.toml`.

image
image

---

Compilation in both Windows and Linux successful after removing `mpris` from the default list.
```
# Linux
cargo install --path ./ --features "mpris"

# Windows
cargo install --path ./
```
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,7 +91,7 @@ objc2 = { version = "0.6", optional = true }
 block2 = { version = "0.6", optional = true }
 
 [features]
-default = ["telemetry", "streaming", "audio-viz-cpal", "mpris", "macos-media", "discord-rpc"]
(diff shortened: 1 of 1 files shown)


LargeModGames:
that was pretty dumb from me, thanks for the fix!

PR #99, opened 2026-02-14
feat: add and remove song from playlist

nthpaul:
# Summary

- adds playlist track management flows in TUI: add track to playlist and remove track from playlist from playlist contexts.
- adds shortcut support for quick add from the currently playing song and existing add/remove actions from track table contexts.
- adds dedicated dialogs for playlist picking (add) and confirmation (remove), including vim-style `h`/`l` navigation for confirmation buttons.
- wires new handler/network events through app state and updates help text/UI affordances for discoverability.
- updates playlist mutation behavior to avoid broad duplicate deletion by applying deterministic single-item removal semantics tied to the selected playlist position.

# Testing

- `/Users/ple/.cargo/bin/cargo fmt --all`
- `/Users/ple/.cargo/bin/cargo test -q network::tests -- --nocapture`
- `/Users/ple/.cargo/bin/cargo test -q`

# Additional notes

- playlist add uses the playlist items API path with `uris` payload.
- playlist remove validates selected row/position against current server data before mutation and then applies ordered playlist replacement in API-sized chunks.
- remove now preserves other duplicate occurrences of the same song by removing only the selected occurrence.
- if a playlist includes local or unavailable items that cannot be represented in mutation payloads, remove reports an actionable status error instead of silently mutating the wrong rows.

### DIFF
--- a/src/network.rs
+++ b/src/network.rs
@@ -2942,16 +2942,105 @@ impl Network {
     format!("playlists/{}/items", playlist_id.id())
   }
 
-  fn add_track_to_playlist_payload(track_uri: &str) -> Value {
-    json!({
-      "uris": [track_uri]
-    })
-  }
-
-  fn replace_playlist_items_payload(uris: &[String]) -> Value {
+  fn playlist_uris_payload(uris: &[String]) -> Value {
     json!({
       "uris": uris
     })
+  }
+
+  fn remove_playlist_item_uri_at_position(
+    mut uris: Vec<String>,
+    position: usize,
+  ) -> anyhow::Result<Vec<String>> {
+    if position >= uris.len() {
+      return Err(anyhow!(
+        "Cannot resolve track position {} in playlist with {} items",
+        position,
+        uris.len()
+      ));
+    }
+    uris.remove(position);
+    Ok(uris)
+  }
+
+  fn playlist_item_uri(item: &PlaylistItem) -> anyhow::Result<String> {
+    match item.track.as_ref() {
+      Some(PlayableItem::Track(track)) => track
+        .id
+        .as_ref()
+        .map(|id| format!("spotify:track:{}", id.id()))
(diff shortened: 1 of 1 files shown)


src/network.rs
```diff
@@ -2904,6 +2924,234 @@ impl Network {
...
+
+    Ok(uris)
+  }
+
+  async fn replace_playlist_with_uris(
+    &self,
+    playlist_id: &PlaylistId<'_>,
+    uris: Vec<String>,
+  ) -> anyhow::Result<()> {
+    let replace_path = format!("playlists/{}/tracks", playlist_id.id());
```
nthpaul: we are mixing /items and /tracks in the same flow here. that can be brittle and harder to maintain.

src/network.rs
```diff
@@ -2904,6 +2924,234 @@ impl Network {
...
+          &path,
+          &[("limit", limit.to_string()), ("offset", offset.to_string())],
+        )
+        .await?;
+
+      for item in page.items {
+        match item.track {
+          Some(PlayableItem::Track(track)) => {
+            let track_id = track
+              .id
```
nthpaul: this fails the whole delete if the playlist has any local or unavailable item. we should skip or preserve those instead of returning an error.

src/network.rs
```diff
@@ -4106,3 +4354,45 @@ fn structurize_playlist_folders(
...
+      payload,
+      json!({
+        "uris": ["spotify:track:6rqhFgbbKwnb9MLmUQDhG6"]
+      })
+    );
+  }
+
+  #[test]
+  fn replace_playlist_payload_uses_uris_shape() {
+    let payload = Network::replace_playlist_items_payload(&[
```
nthpaul: these tests only check payload shape. we should add a test for the single-occurrence removal algorithm (remove one index and keep order).

LargeModGames:
This is amazing tysm!

PR #102, opened 2026-02-16
Implement new logging system

### LINKED ISSUE #103: Implement logging system for spotatui

As per my understanding, you'll only access the logs only if the program panics. There are various GitHub issues that won't cause a panic, hence it's hard to diagnose.

https://github.com/LargeModGames/spotatui/issues/96#issuecomment-3905022507

> [@LargeModGames](https://github.com/LargeModGames) I advise you to add proper logging/debugging to see what happens after the program exits.
> 
> I think this will solve the issue.
> 
> If you want I can look into it when I get home later and maybe make a pr.

Use a library like log to write log messages into a file to watch while using spotatui

**Additional context**

I made changes to the code, I've a PR linked to this issue.

AFE123xr:
# Summary

I modified spotatui to use log and fern for logging. Instead of waiting for panics to occur, you can watch the logs in `/tmp/spotatui_logs/spotatuilog{PID}` while using spotatui

closes #103 

# Testing

You can run spotatui, and run `tail -f /tmp/spotatui_logs/{PID}` while running spotatui. PID is the Process ID of the current spotatui session. You can also check the file with less and use grep to isolate errors. 

# Additional notes

image

I'm watching the logs while running spotatui.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1329,6 +1329,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "fern"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4316185f709b23713e41e3195f90edef7fb00c3ed4adc79769cf09cc762a3b29"
+dependencies = [
+ "log",
+]
+
+[[package]]
 name = "fiat-crypto"
 version = "0.2.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4644,6 +4653,7 @@ dependencies = [
  "crossterm",
  "dirs",
  "discord-rich-presence",
+ "fern",
  "futures",
  "librespot-connect",
  "librespot-core",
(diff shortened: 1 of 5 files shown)


LargeModGames:
@all-contributors please add @AFE123x for code, bug

topongo:
i've seen the new feature, kinda neat, not gonna lie. my only problem is during debugging: it could be a lot more easy if we could use a single log file in append mode.

but i could already see a problem: what if two instances are open simultaneously?

AFE123xr:
> but i could already see a problem: what if two instances are open simultaneously?
>

The log file is named after the program's PID. If there are two instances, there'll be two files with different PID.  The TUI also mentions where the log file is located for the specific instance

image

>
> i've seen the new feature, kinda neat, not gonna lie. my only problem is during debugging: it could be a lot more easy if we could use a single log file in append mode.
> 

Thanks! I feel like it'd be confusing to keep track of which process the log statement is coming from if we were to use a single log file.

topongo:
> The TUI also mentions where the log file is located for the specific instance

oh i didn't even notice, thanks. but what i meant to say is that during normal developement it could be annoying to continuosly change the file i'm `tail -f`ing from.

it could be nice to use a command line argument and/or an environment variable to tell the logger to write (append/trunc) to a specific file

PR #120, opened 2026-02-26
Issue/10 macos media keys

### LINKED ISSUE #10: Support media keys in Macos
It would be nice if spotatui could be controlled using the media keys. That way I wouldn't have to focus the tui to change songs, pause etc.

Vi1i:
# Summary
Fixes #10  — macOS media keys (play/pause, next, previous) now control spotatui via MPRemoteCommandCenter, and the app appears in Control Center's Now Playing widget.

Three root causes prevented the existing macOS media integration from working:
- **No NSApplication initialized** — MPRemoteCommandCenter requires one to route events; without it, macOS falls through to Music.app
- **No NSRunLoop ticking** — macOS delivers events via the run loop; the original code never ticked it (both media thread and main thread run loops are required)
- **No Now Playing metadata published** — SetMetadata was a TODO no-op, so macOS didn't consider the app a media player

# Changes
- `src/infra/macos_media.rs`: Initialize NSApplication with Prohibited activation policy, interleave NSRunLoop ticks with command processing via `tokio::select!`, implement `handle_now_playing_command` for metadata/playback state/position
- `src/main.rs`: Wire `MacMediaManager` into `handle_player_events` (Playing, Paused, Seeked, TrackChanged, Stopped, EndOfTrack), add main-thread NSRunLoop tick in Event::Tick handler, restore `lock().await` for TrackChanged to match MPRIS version

# Test
- [x] Build with `--features macos-media`
- [x] Verify media keys control playback (play/pause, next, previous)
- [x] Verify app appears in Control Center Now Playing widget with track metadata
- [x] cargo fmt --all
- [x] cargo clippy --locked -- -D warnings
  - Fixed issues introduced by these changes, not pre-existing ones
- [x] cargo test --locked

### DIFF
--- a/src/infra/macos_media.rs
+++ b/src/infra/macos_media.rs
@@ -38,7 +38,7 @@ pub enum MacMediaEvent {
 
 /// Commands to send TO the Now Playing center to update its state
 #[derive(Debug, Clone)]
-#[allow(dead_code)]
+#[allow(dead_code, clippy::enum_variant_names)]
 pub enum MacMediaCommand {
   SetMetadata {
     title: String,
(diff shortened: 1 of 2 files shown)


Vi1i:
The failing check, https://github.com/LargeModGames/spotatui/actions/runs/22449893702/job/65018872365?pr=120, is failing on Clippy errors from other sources.

LargeModGames:
@all-contributors add @Vi1i for code, bug

PR #149, opened 2026-03-05
chore: update nix flake

Kathund:
# Summary

This change makes it so that it loads the commit from the user's flake.lock
This means that it will build of the commit that the user is on. The user can use `nix flake update spotatui` to update the commit they are on and when they next rebuild they will be put on the new version

This is relating to https://github.com/LargeModGames/spotatui/discussions/143#discussioncomment-15998939

# Testing

# Additional notes

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
         packages = rec {
           default = pkgs.rustPlatform.buildRustPackage {
             pname = "spotatui";
-            version = "0.37.2";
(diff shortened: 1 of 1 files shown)


kalebris:
not sure if you are using mpris, but maybe mpris enablment could be incorporated. I have worked around this here: https://github.com/LargeModGames/spotatui/issues/150

Kathund:
> not sure if you are using mpris, but maybe mpris enablment could be incorporated. I have worked around this here: #150

Hmm interesting. I don't think this should be in the scope of this pr but I will look into making a followup pr if I have time

LargeModGames:
@all-contributors add @Kathund for platform

PR #171, opened 2026-03-17
Void Linux installation guide

eiseq:
Added commands to install spotatui on Void Linux.

It uses unofficial package repo ([Blackhole-vl](https://github.com/Event-Horizon-VL/blackhole-vl))


LargeModGames:
@all-contributors add @eiseq for doc

PR #172, opened 2026-03-18
feat(mpris): decouple from streaming and add full state sync

higorprado:
## Problem

MPRIS was gated behind the `streaming` feature flag and only initialized when a native librespot player was active. Users on external Spotify devices (phone, spotifyd, etc.) got no MPRIS support — `playerctl -l` would not list spotatui.

State sync was also incomplete: playback status, shuffle and repeat were only pushed via librespot player events, so external clients (playerctl, KDE widget, waybar) saw stale state on API-based playback.

## Changes

- Remove `streaming` from the `mpris` feature; add `mpris` to defaults. All code stays behind the existing Linux guard.
- Initialize MPRIS unconditionally instead of guarding behind `streaming_player.is_some()`
- Add IoEvent fallback in `handle_mpris_events` for every control action when no native streaming player is present
- Replace `update_mpris_metadata` with `update_mpris_state`, which also syncs playback status, shuffle and repeat each tick
- Consolidate the two duplicated `handle_player_events` into one

CI build (`--no-default-features --features telemetry`) is unaffected.

### DIFF
--- a/src/core/app.rs
+++ b/src/core/app.rs
@@ -3518,8 +3518,8 @@ impl App {
 
 #[cfg(test)]
 mod tests {
-  use crate::core::test_helpers::{private_user, simplified_playlist};
   use super::*;
+  use crate::core::test_helpers::{private_user, simplified_playlist};
   use chrono::{Duration as ChronoDuration, Utc};
   use rspotify::model::{artist::SimplifiedArtist, idtypes::PlaylistId};
   use rspotify::prelude::Id;


LargeModGames:
That was pretty dumb of me, idk why i didnt think about that.

Did you also test manually that mpris controls now work with external devices?

higorprado:
> Did you also test manually that mpris controls now work with external devices?

Yes! Tested with an Android phone as the active Spotify device. I didn't do exhaustive testing across all edge cases, but the core controls are functional.

LargeModGames (approved):

PR #200, opened 2026-04-01
Greybeard mode

benalleng:
I did not expect mouse inputs to be accepted in the tui and when I clicked the window to switch focus it immediately played a new song. I think it would be nice to have a mode where mouse inputs are not accepted.

# Summary

Add a new setting in behaviors to disable mouse inputs entirely

# Testing

ran formatting linting and testing before commiting

# Additional notes

Also added a display change such that when this mode is active the UI does not show "Click" as the open settings hint but instead just stringifies the current open setting keybind.  The second commit tries to simplify some of the settings display code for help and settings there further but that is more opinionated as it removes "Click" all together. happy to talk more about whether to include that commit or not.

image

### DIFF
--- a/src/tui/ui/search.rs
+++ b/src/tui/ui/search.rs
@@ -16,8 +16,8 @@ use super::util::{
 };
 
 const COMPACT_TOP_ROW_THRESHOLD: u16 = 60;
-const COMPACT_HELP_WIDTH: u16 = 8;
-const COMPACT_SETTINGS_WIDTH: u16 = 12;
+const COMPACT_HELP_WIDTH: u16 = 6;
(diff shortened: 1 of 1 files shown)


LargeModGames:
Nice changes, thank you! I'll merge it tomorrow when I wake up.

Why did you call it Greybeard mode tho? I have never heard of that term before

benalleng:
It is a reference to the old men in the computer world that love their old ways, and you'd have to be pretty old at this point to be used to a computer without a [mouse](https://en.wikipedia.org/wiki/Xerox_Star)

LargeModGames:
@all-contributors add @benalleng for code

PR #216, opened 2026-04-15
fix: preserve refresh token to prevent startup reauthentication loop

shilicioo:
Fixes #211

# Summary

**Startup reauthentication loop**: Fixed forced Spotify reauthentication on nearly every launch by (1) preserving refresh_token when saving updated access tokens and (2) explicitly refreshing expired tokens before validation instead of forcing re-authentication

# Testing

- cargo fmt --all
- cargo clippy --locked -- -D warnings
- cargo test --locked

Manual testing:
- after authenticating, change "expires_at" to date/time in the past (location: ~/.config/spotatui/.spotify_token_cache*.json)
- re-launch spotatui app, expected behavior: no auth screen, app launches without interruptions
- verify spotify_token_cache*.json "expires_at" is updated and is in future again
- observe startup log, it should contain entries "cached token is expired, attempting automatic refresh" and "token refreshed successfully"

PR #217, opened 2026-04-15
Fix re-login on every reboot (PKCE refresh token not persisted)

### LINKED ISSUE #211: Constant need for reauth every startup
**Describe the bug**
Almost every time I open Spotatui it asks for me to reauth with Spotify

**Expected behavior**
Boot straight to library

**Screenshots**

Image

 - OS: Nobara Linux
 - Terminal: kitty
 - Version  0.38.0

_][15:55:05][spotatui][INFO] spotatui 0.38.0 starting up
[2026-04-11][15:55:05][spotatui][INFO] audio backend initialized
[2026-04-11][15:55:05][spotatui][INFO] panic hook configured
[2026-04-11][15:55:05][spotatui][INFO] user config loaded successfully
[2026-04-11][15:55:06][spotatui][INFO] client authentication config loaded
[2026-04-11][15:55:06][spotatui][INFO] authentication token loaded from cache
[2026-04-11][15:55:06][rspotify_http::reqwest][INFO] Making request RequestBuilder { method: POST, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("accounts.spotify.com")), port: None, path: "/api/token", query: None, fragment: None }, headers: {"content-type": "application/x-www-form-urlencoded"} }
[2026-04-11][15:55:06][spotatui][INFO] cached authentication token is invalid, re-authentication required
[2026-04-11][15:55:06][spotatui][INFO] starting spotify authentication flow on port 8989
[2026-04-11][15:55:06][rspotify::auth_code_pkce][INFO] Building auth URL
[2026-04-11][15:55:06][rspotify::auth_code_pkce][INFO] Generating PKCE codes
[2026-04-11][15:55:41][spotatui][INFO] authorization code received, requesting access token
[2026-04-11][15:55:41][rspotify::auth_code_pkce][INFO] Requesting PKCE Auth Code token
[2026-04-11][15:55:41][rspotify_http::reqwest][INFO] Making request RequestBuilder { method: POST, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("accounts.spotify.com")), port: None, path: "/api/token", query: None, fragment: None }, headers: {"content-type": "application/x-www-form-urlencoded"} }
[2026-04-11][15:55:41][rspotify::clients::base][INFO] Token cache write ignored (not configured)
[2026-04-11][15:55:41][spotatui][INFO] token cached to /home/paper/.config/spotatui/.spotify_token_cache_d420a117.json
[2026-04-11][15:55:41][spotatui][INFO] successfully authenticated with spotify
[2026-04-11][15:55:41][spotatui][INFO] Using ncspot shared client ID. If it breaks in the future, configure fallback_client_id in client.yml.
[2026-04-11][15:55:41][spotatui][INFO] app state initialized_

F1xxs:
Fixes #211 

Spotify's PKCE flow rotates the refresh token on every use, each refresh token is single-use and the old one is immediately invalidated. The app was never writing the updated token back to disk, so every reboot would fail to authenticate with the stale on-disk token.

Three related issues were fixed:

- Startup: rspotify silently auto-refreshes an expired token during the startup validation probe (`spotify.me())`, rotating the refresh token in memory without saving. The token is now written to disk immediately after `ensure_auth_token` completes.

- Wrong expiry tracking: `app.spotify_token_expiry` was computed as now + expires_in, where `expires_in` is always 3600 regardless of when the token was issued. This caused `refresh_authentication()` to fire too late, leaving rspotify's internal `auto_reauth` to handle mid-session refreshes, which also rotated the token without saving. Fixed by using `expires_at` (the absolute timestamp stored in the token) instead.

- Runtime refresh: `refresh_authentication()` was updating the token in memory but never writing it to disk. The `Network` struct now holds the token cache path and saves the full token (including the new refresh token) after every successful refresh.


LargeModGames:
I havent really looked at the code yet, plan to do that tomorrow, but what is the difference between this PR and #216?

shilicioo:
> I havent really looked at the code yet, plan to do that tomorrow, but what is the difference between this PR and #216?

there are few core differences:
- this one relies on rspotify token refresh, while #216 uses manual validation and refresh
- #217 has cleaner architecture with network struct enhancement

if all the edge-cases were tested, I like #217 more :)
(+ add test coverage)

F1xxs:
I have tested the code with:
- cargo fmt --all
- cargo test --locked

I've also now tried testing with @shilicioo tests from #216 and only one test failed:
```bash
thread 'tests::test_save_token_preserves_refresh_token_when_missing' (35153) panicked at src/main.rs:3147:5:
assertion `left == right` failed


test result: FAILED. 162 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.25s
error: test failed, to rerun pass `-p spotatui --bin spotatui```

**Comment by LargeModGames** [comment]
Looked through this one a bit more too and I think it still needs one important change.

I like the idea of saving the token after startup auth/refresh, but we still need to preserve the old `refresh_token` before writing the new token to disk.

Spotify does not always send back a new `refresh_token` on refresh, so in that case this can save `refresh_token: null` and then we are right back in the browser login flow on next launch.

So I think this needs the same "keep the existing refresh token if the new one does not have one" behavior, plus a test for that case.

**Comment by F1xxs** [comment]
Added preservation of the old `refresh_token` when the new one is absent. This is applied in both `save_token_to_file` and the inline save inside `refresh_authentication()` , so neither path can overwrite a valid refresh token with null.

Also borrowed the relevant test cases from @shilicioo's PR #216.

File: src/main.rs [hunk 1]
```diff
@@ -471,8 +471,15 @@ fn update_macos_metadata(
...
-  if let Some(ref token) = *token_lock {
-    let token_json = serde_json::to_string_pretty(token)?;
+  if let Some(mut token) = (*token_lock).clone() {
+    if token.refresh_token.is_none() && path.exists() {
+      if let Ok(old_json) = fs::read_to_string(path) {
+        if let Ok(old_token) = serde_json::from_str::<Token>(&old_json) {
+          token.refresh_token = old_token.refresh_token;
+        }
+      }
+    }
```
LargeModGames: We preserve the previous refresh_token only in the cloned token we serialize here,
but spotify.token in memory still ends up with refresh_token = None when Spotify omits it on refresh. That means the current session can still fail its next refresh even if the cache file is correct. Can we also write the recovered refresh token back into the in-memory client state here? I think Network::refresh_authentication() needs the same fix

src/main.rs
```diff
@@ -1139,14 +1146,34 @@ of the app. Beware that this comes at a CPU cost!",
...
+      SystemTime::UNIX_EPOCH + std::time::Duration::from_secs(unix_secs)
+    } else {
+      let expires_in_secs = token.expires_in.num_seconds().max(0) as u64;
+      SystemTime::now()
+        .checked_add(std::time::Duration::from_secs(expires_in_secs))
+        .unwrap_or_else(SystemTime::now)
+    }
   } else {
     drop(token_lock);
     return Err(anyhow!("Authentication failed: no valid token available"));
```
LargeModGames, line 1180: This makes startup depend on being able to write the cache file. If ensure_auth_token() already loaded a valid cached token, a cache write failure here
seems like it should be non-fatal so the app can still run for the current session. Could we handle this like refresh_authentication() and keep cache-save failures as warnings instead of aborting startup?

LargeModGames (changes requested):

F1xxs:
Thanks for the detailed review. Both points addressed:
- In-memory refresh_token not updated
  Fixed in both `save_token_to_file` and `refresh_authentication`. Previously we cloned
  the token, recovered the `refresh_token` into the clone, and wrote only the clone to
  disk. The in-memory `spotify.token` still had `refresh_token = None`. Now both
  functions take a mut lock guard and use ref mut token on the live `Option<Token>`, so
  the recovered `refresh_token` is written back through the lock into the in-memory
  client state at the same time as it's persisted to disk.

- Startup cache write failure should be non-fatal
  Changed `save_token_to_file(...).await?` to an `if let Err(e) block that logs with
  log::warn!`. A cache write failure at startup (e.g. read-only
  filesystem) no longer aborts the session. The app runs on the valid in-memory token
  it already has.

I have tested the code with:

- cargo fmt --all
- cargo test --locked

LargeModGames (approved):

LargeModGames:
This looks good to me now, thank you very much!

Issue #231, opened 2026-04-29
[Bug] Currently selected theme presets and colours aren't properly updated and displayed on UI


Moaht:
Describe the bug
Currently selected theme preset does not show up as selected when reopening the theme page in settings after it is saved.
Further to this, when cycling through the presets, the RGB values are not updated. We just see whatever was loaded when opening the settings.

To Reproduce
Steps to reproduce the behaviour:

1. Go to Settings > themes
2. Change theme to something different from Cyan such as Spotify
3. Save and exit
4. Go to settings > themes
5. Cyan theme still shows up as the displayed option even if it is not selected.

1. Go to Settings > themes
2. Press enter to cycle through the presets
3. The RGB values do not update

Expected behaviour
At Step 5, the currently in use theme should be seen right away when entering the themes section in settings. This is an indicator that it is the currently selected theme. 
At Step 3, the currently focused theme preset values should be displayed when pressing enter to switch between the options.

Screenshots
n/a

Desktop (please complete the following information):

    OS: [ Linux ]
    Terminal [ gnome-terminal, tmux]
    Version [ 0.38.1]

Additional context

Issue #230, opened 2026-04-29
[Bug] Changing the colours in themes does not actively change the theme colours


Moaht:
Describe the bug
Changing the RGB values of the theme does not do anything, it simply displays the values temporarily and then discards them.

To Reproduce
Steps to reproduce the behaviour:

1. Go to settings > themes
2. Select a theme colour key to change
3. Change the value
4. Save and exit
5. Check the UI
6. The colours have not changed

Expected behaviour
When you change the colour values, you should be able to save it. This should be loaded into a 'custom' preset that gets updated whenever the values are changed. When it is saved, the 'custom' preset is selected.

Screenshots

Desktop (please complete the following information):

    OS: [Linux ]
    Terminal [ gnome-terminal, tmux ]
    Version [ 0.38.1 ]

Additional context

PR #232, opened 2026-04-29
Fix: Themes preset editing and persistence and display of theme item values

Moaht:
# Summary

This pull request addresses the issues in #230 and #231.

It adds a new theme preset called Custom which gets activated whenever the RGB values are edited. This means that editing the RGB values of a theme item now has an effect.  After editing a preset, the currently selected preset changes to Custom. The Custom setting becomes populated with the values of the edited preset along with the edits that make it customised. Whenever the user cycles through the presets, they can reload the preset and then come back to the Custom settings they changed earlier. This is saved using the existing values in the config.
Currently selected presets, not just the RGB values for the items, are now tracked in the config. 
Cycling through the theme presets when pressing enter now shows the RGB values that correspond to the in focus preset. 
Exiting the themes settings and re-entering, the active preset is seen right away which helps the user understand that it is in use. 

# Testing
Running `cargo fmt --all` resulted in no noise.

Running `cargo clippy --locked -- -D warnings` resulted in:
```Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.36s```

Running `cargo test --locked` resulted in: 
```Resume this session with:                                   │test result: ok. 165 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s```

# Additional notes
Screenshot From 2026-04-29 20-00-14

### DIFF
--- a/src/core/app.rs
+++ b/src/core/app.rs
@@ -3737,6 +3737,28 @@ impl App {
       }
     }
   }
+
+  /// Updates the colour RGB entries when switching through the presets in themes
+  pub fn sync_theme_color_settings(&mut self, theme: &crate::core::user_config::Theme) {
+    let mappings: [(&str, ratatui::style::Color); 11] = [
+      ("theme.active", theme.active),
+      ("theme.banner", theme.banner),
+      ("theme.hint", theme.hint),
+      ("theme.hovered", theme.hovered),
+      ("theme.selected", theme.selected),
+      ("theme.inactive", theme.inactive),
+      ("theme.text", theme.text),
+      ("theme.error_text", theme.error_text),
+      ("theme.playbar_background", theme.playbar_background),
+      ("theme.playbar_progress", theme.playbar_progress),
+      ("theme.highlighted_lyrics", theme.highlighted_lyrics),
+    ];
+    for setting in &mut self.settings_items {
+      if let Some((_, color)) = mappings.iter().find(|(id, _)| *id == setting.id) {
+        setting.value = SettingValue::Color(color_to_string(*color));
+      }
+    }
+  }
 }
 
 #[cfg(test)]
--- a/src/core/app.rs
+++ b/src/core/app.rs
@@ -3280,6 +3280,12 @@ impl App {
             value: SettingValue::Color(color_to_string(self.user_config.theme.error_text)),
           },
           SettingItem {
+            id: "theme.error_border".to_string(),
+            name: "Error Border Color".to_string(),
+            description: "Border color for error messages".to_string(),
+            value: SettingValue::Color(color_to_string(self.user_config.theme.error_border)),
+          },
+          SettingItem {
             id: "theme.playbar_background".to_string(),
             name: "Playbar Background".to_string(),
             description: "Background color for playbar".to_string(),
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by LargeModGames** [comment]
Could you check if the old preset themes still work? Im not sure how backward compatibility for existing configs that have custom theme colors but no theme.preset yet work

**Comment by Moaht** [comment]
> Could you check if the old preset themes still work? Im not sure how backward compatibility for existing configs that have custom theme colors but no theme.preset yet work

Everything in the config should works the same as it did before except now it has a "preset" entry. 

The other keys in the config are exactly the same and are used for the Custom preset.

Actually, just checking it through now: it loads the presets fine and the custom settings are saved in the session, but there seems to be an issue with the custom values being loaded after a reboot. I'll have to take a look and fix that. 
When I'm done I'll try loading up custom settings in the config in the current version and then seeing if they are compatible with the changes, i.e they persist and work after the update (is that what you meant)?

**Comment by LargeModGames** [comment]
> Actually, just checking it through now: it loads the presets fine and the custom settings are saved in the session, but there seems to be an issue with the custom values being loaded after a reboot. I'll have to take a look and fix that. 
> When I'm done I'll try loading up custom settings in the config in the current version and then seeing if they are compatible with the changes, i.e they persist and work after the update (is that what you meant)?

yes that is what i meant

**Comment by Moaht** [comment]
> > Actually, just checking it through now: it loads the presets fine and the custom settings are saved in the session, but there seems to be an issue with the custom values being loaded after a reboot. I'll have to take a look and fix that.
> > When I'm done I'll try loading up custom settings in the config in the current version and then seeing if they are compatible with the changes, i.e they persist and work after the update (is that what you meant)?
> 
> yes that is what i meant

Okay. Then yes it all works exactly as it should. 

A couple of things were holding things up:

I thought there was some issue with the loading of the background colours but it was just the "Reset" value which uses the set terminal colours. Obviously this can vary quite a bit tmux,screen,tty vim etc. 

In b972a97 I've now added new options in the theme menu for users to change the other settings such as background colour etc if they wish. And you can still reset to terminal default using "Reset" at any time in the UI or in the config. 

After upgrading to f04462e, configs would not have their past custom colours from their configs loaded immediately. They would have to go to Settings > Themes and select Custom and their custom colours from before would then be shown. I've now changed this in 64890de so that a simple check to see if the preset key doesn't exist and any of the colour values exist in the config, to use the custom config option. This now solves the issue of backwards compatibility. 

Running `cargo fmt --all` resulted in no noise.

Running `cargo clippy --locked -- -D warnings` resulted in:
```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.11s                                                                                                                                            
```

Running `cargo test --locked` resulted in:
```
Finished `test` profile [unoptimized + debuginfo] target(s) in 2.93s                    
Running unittests src/main.rs (target/debug/deps/spotatui-c425a8bb622d87c9)                                                                                                                                   
test result: ok. 165 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.10s
```

LargeModGames (approved):

LargeModGames:
@all-contributors add @Moaht for code

Issue #238, opened 2026-05-09
UI Freezes while trying to load Recommended Songs


rlpratyoosh:
**Describe the bug**
When trying to load recommended songs from the search list, the UI freezes for about 10 seconds.

**To Reproduce**
1. Search a song
2. Select a song
3. Press `r`
4. UI Freezes

**Expected behavior**
UI should still be usable and it shouldn't take about 10 seconds to load recommended songs.

**Desktop**
 - OS: Arch Linux
 - Terminal: Simple Terminal
 - Version: 0.9.3

**Additional context**
I have written a fix for it and would open a PR linking to the issue!

PR #239, opened 2026-05-09
Fix UI freeze

rlpratyoosh:
# Summary
Fixes Issue #238  

- Taking app lock before fetching recommended songs was freezing the UI, so moved it after the fetch.
- Sequentially fetching tracks was causing **N+1** problem resulting in approx 10s latency. Using `tracks` endpoint was reducing the latency from 10s to 2s. But since it was deprecated, I used `future::join_all()` to fetch tracks concurrently, reducing the *10s** latency to approx **4s**.

# Testing

```bash
cargo fmt --all
cargo clippy --locked -- -D warnings
cargo test --locked
```
```bash
test result: ok. 174 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.10s
```

### DIFF
--- a/src/infra/network/recommend.rs
+++ b/src/infra/network/recommend.rs
@@ -1,6 +1,7 @@
 use super::Network;
 use crate::core::app::{ActiveBlock, RouteId, TrackTableContext};
 use anyhow::anyhow;
+use futures::future;
 use rspotify::model::{
   enums::Country,
   idtypes::{ArtistId, TrackId},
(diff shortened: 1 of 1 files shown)


rlpratyoosh:
Apologies for the CI trip-up... I split the work into atomic commits for easier review and accidentally dropped the changes made by `cargo fmt`. I just amended the commit and force-pushed the corrected formatting.

LargeModGames:
@all-contributors add @rlpratyoosh for code, bug

PR #257, opened 2026-05-16
fix token refresh handling

### LINKED ISSUE #256: Auth token refresh does not work
**Describe the bug**
Auth token refresh fails leading to force manual re-log-in

**To Reproduce**
1. Log in in spotatui
2. Leave the session idle for 30+ min
3. Try to interact with spotatui

**Expected behavior**

 - OS: macOS Tahoe v.26.4.1
 - Terminal: kitty
 - Version 0.38.2

shilicioo:
Fixes #256 

# Summary
- Disabled rspotify auto-refresh so spotatui owns both refresh and persistence
- Added a shared locked refresh helper with 60s safety margin and cache persistence
- Added UI-side refresh coalescing to stop blocking UI in a loop

# Testing
cargo fmt --all
cargo clippy --no-default-features --features telemetry -- -D warnings
cargo test --no-default-features --features telemetry


LargeModGames (changes requested):
Disabling rspotify's built in refresh here is only safe if every spotify call gets migrated to the new refresh and cache path. There are other existing paths that still call rspotify directly or use the legact request handler (for example search, metadata, user and playback helpers). So when those requests get a 401 because the access token was revoked, skewed or races expiry they no longer refresh/retry as before so the user action fails.

Tihs could still work but would require a lot of work including:
- no direct authenticated self.spotify.*().await API calls remain except auth/token specific calls
- all request helpers share the same refresh-and-retry behaviour 
- refreshed tokens are persisted without dropping the refresh token
- 401 retry is covered by tests. ideally for the shared heler insteand of every endpoint

I know this is a lot of work to do yourself so if you need any help just let me know and ill help. Thanks a lot for contributing!

shilicioo:
@LargeModGames 
Good point🙂 
I’d split this into two steps then. To first fix the blocking issue as this makes the app barely usable: when I come back to spotatui after 1hr meeting - I go through Ctrl+C to kill it + re-auth, that's painful 🥲 

For this PR, I’d keep rspotify auto-refresh enabled for now, so existing direct self.spotify.*().await calls continue to behave as before. On top of that, we can persist refreshed tokens through spotatui’s cache path via the rspotify token callback, preserve the cached `refresh_token` when Spotify does not return a new one, and keep the explicit pre-expiry refresh/coalescing logic to avoid the relogin loop (theoretically, I need to test this properly😄)

Then second, I’ll create a separate branch for the proper cleanup: move direct rspotify calls behind shared request helpers, centralise 401 refresh/retry handling, and only then remove reliance on rspotify refresh completely.

That way we unblock users quickly, and then focus on more significant refactoring.

Let me know what you think

LargeModGames:
> @LargeModGames Good point🙂 I’d split this into two steps then. To first fix the blocking issue as this makes the app barely usable: when I come back to spotatui after 1hr meeting - I go through Ctrl+C to kill it + re-auth, that's painful 🥲
> 
> For this PR, I’d keep rspotify auto-refresh enabled for now, so existing direct self.spotify.*().await calls continue to behave as before. On top of that, we can persist refreshed tokens through spotatui’s cache path via the rspotify token callback, preserve the cached `refresh_token` when Spotify does not return a new one, and keep the explicit pre-expiry refresh/coalescing logic to avoid the relogin loop (theoretically, I need to test this properly😄)
> 
> Then second, I’ll create a separate branch for the proper cleanup: move direct rspotify calls behind shared request helpers, centralise 401 refresh/retry handling, and only then remove reliance on rspotify refresh completely.
> 
> That way we unblock users quickly, and then focus on more significant refactoring.
> 
> Let me know what you think

Yes, that sounds like the right split.

  For this PR, please keep `token_refreshing` enabled so existing direct
  `self.spotify.*().await` paths keep their current refresh behavior. Then focus this PR
  on persisting refreshed tokens via rspotify’s token callback, preserving the existing
  cached `refresh_token` when Spotify does not return a new one, and keeping the pre-
  expiry refresh/coalescing logic if it stays small.

  One important thing: if the pre-event refresh path stays, it must clear
  `app.is_loading` on refresh failure before returning. Otherwise normal actions can
  leave the UI stuck loading.

  The larger cleanup of routing every authenticated request through shared helpers with
  centralized 401 refresh/retry should be a separate PR.

shilicioo:
> > @LargeModGames Good point🙂 I’d split this into two steps then. To first fix the blocking issue as this makes the app barely usable: when I come back to spotatui after 1hr meeting - I go through Ctrl+C to kill it + re-auth, that's painful 🥲
> > For this PR, I’d keep rspotify auto-refresh enabled for now, so existing direct self.spotify.*().await calls continue to behave as before. On top of that, we can persist refreshed tokens through spotatui’s cache path via the rspotify token callback, preserve the cached `refresh_token` when Spotify does not return a new one, and keep the explicit pre-expiry refresh/coalescing logic to avoid the relogin loop (theoretically, I need to test this properly😄)
> > Then second, I’ll create a separate branch for the proper cleanup: move direct rspotify calls behind shared request helpers, centralise 401 refresh/retry handling, and only then remove reliance on rspotify refresh completely.
> > That way we unblock users quickly, and then focus on more significant refactoring.
> > Let me know what you think
> 
> Yes, that sounds like the right split.
> 
> For this PR, please keep `token_refreshing` enabled so existing direct `self.spotify.*().await` paths keep their current refresh behavior. Then focus this PR on persisting refreshed tokens via rspotify’s token callback, preserving the existing cached `refresh_token` when Spotify does not return a new one, and keeping the pre- expiry refresh/coalescing logic if it stays small.
> 
> One important thing: if the pre-event refresh path stays, it must clear `app.is_loading` on refresh failure before returning. Otherwise normal actions can leave the UI stuck loading.
> 
> The larger cleanup of routing every authenticated request through shared helpers with centralized 401 refresh/retry should be a separate PR.

added the discussed scope of initial fix, please have a look

LargeModGames (approved):

PR #273, opened 2026-05-25
Auto-select native streaming player

eduardolabo:
## Description

This PR resolves the `404 Not Found` error returned by the Spotify API when starting playback with no active device selected on the account.

When native streaming is enabled and the player is connected, the application will now automatically fallback to and activate the native player as the default playback target if no other active device is detected.

### 1. Playback Auto-Select Logic
* Extracted the device selection state checks into a new helper function `should_fallback_to_native_playback` in `src/infra/network/playback.rs`.
* Updated `start_playback` to automatically trigger a transfer and activation of the native player when it is connected and no other active playback device is detected.

### 2. Unit Testing
* Added `test_should_fallback_to_native_playback` in `src/infra/network/playback.rs` covering all key scenarios:
  * Native streaming is already active -> `true` (maintains playback).
  * No active device, but native player is connected -> `true` (auto-fallback).
  * Another device is already active -> `false` (avoids hijacking active playback).
  * No active device, but native player is disconnected -> `false` (no-op).

### Automated Tests
Ran the full test suite (including streaming configurations):
```bash
cargo test

### DIFF
--- a/src/infra/network/playback.rs
+++ b/src/infra/network/playback.rs
@@ -525,7 +525,26 @@ impl PlaybackNetwork for Network {
 
     // Check if we should use native streaming for playback
     #[cfg(feature = "streaming")]
-    if is_native_streaming_active_for_playback(self).await {
+    let use_native = {
+      let is_active = is_native_streaming_active_for_playback(self).await;
+      if is_active {
+        true
+      } else {
+        let app = self.app.lock().await;
+        let has_active_device = app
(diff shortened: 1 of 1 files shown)


eduardolabo:
I haven't tested this enough, closing for now

LargeModGames:
> I haven't tested this enough, closing for now

feel free to open it again when you're ready

PR #276, opened 2026-05-27
Refactor auth token refresh and retry handling

shilicioo:
# Summary

Refactors Spotify auth handling so authenticated API calls go through a shared refresh/cache/retry path instead of relying on rspotify auto-refresh. This prevents expired or refreshed tokens from breaking the cache and forcing re-login.

# Testing
  - `cargo fmt --all`
  - `cargo check`

# Additional notes
No UI changes.

This also preserves refreshed tokens without dropping the cached `refresh_token`, disables rspotify auto-refresh after moving authenticated calls to shared helpers, and adds shared 401 refresh/retry coverage.

P.S.
I'd suggest someone does another iteration of manual testing, I did and haven't faced any issues, but to make sure I didn't miss anything

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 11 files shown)


LargeModGames:
have been trying it for about an hour now and havent found any real issues. Will keep using it for a bit more but so far it looks good. 

This really is amazing work thank you so much!

PR #291, opened 2026-06-08
System Media Tansport Controls Integration for Windows

### LINKED ISSUE #229: Media Key/RPC Server Support
Being able to control the playback of the native streaming using mouse/keyboard buttons or via a RPC server would be useful when running out of focus. 

Super ideal is support for media keys via MPRIS/SMTC for native OS integration. 

**Describe alternatives you've considered**
An togglable RPC server that can receive playback commands would be an ok alternative as platform-specific bridge software could be created. 

**Additional context**
I just wanna use my keyboard/mouse buttons man lol.

rlpratyoosh:
# Summary

- Added `smtc-tokio` library for SMTC integration in the app
- Pressing media keys to control playback is now supported in windows.
- Resolves #229 

# Testing

- cargo fmt --all
- cargo test --locked (244 passed)

# Additional notes

https://github.com/user-attachments/assets/e9898ff3-a87b-4af8-86ed-df390512437d

- `smtc-tokio` library is created by me, and can be checked on https://github.com/rlpratyoosh/smtc-tokio

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- **SMTC Integration**: System Media Transport Controls is now integrated with the app for Windows users. Users can now control playback state using media keys and check playback state in media flyouts ([#229](https://github.com/LargeModGames/spotatui/issues/229)).
 - **Click and drag to seek on the playbar**: The progress bar is now interactive. Click anywhere on the gauge to jump to that position, or click and drag to scrub. Control buttons keep priority, the time label stays non-clickable, and seeks reuse the existing native and throttled-API paths ([#157](https://github.com/LargeModGames/spotatui/issues/157)).
 
 ### Fixed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5443,6 +5443,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67b1b7a3b5fe4f1376887184045fcf45c69e92af734b7aaddc05fb777b6fbd03"
 
 [[package]]
+name = "smtc-tokio"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "efdb8287e392ee9fe68f7dbf87d5673e6ed465f0aaab99b0e6680c346fdc1e6e"
+dependencies = [
+ "tokio",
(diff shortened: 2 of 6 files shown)


LargeModGames:
Thanks for this. Really clean integration, and it mirrors the existing macOS (`macos-media`) wiring closely, which makes it easy to follow. A few things before merge:

**Should fix:**
- `windows-media = ["smtc-tokio"]` doesn't pull in `streaming`, but the SMTC manager setup in `runtime.rs` references `streaming_player`, which only exists under the `streaming` feature. Building `--features windows-media` without `streaming` won't compile. Please make it `windows-media = ["smtc-tokio", "streaming"]` to match `macos-media`. Default builds are fine since they include both; this is just so the feature is self-contained.

**Worth a look (non-blocking):**
- Both the player-event handler (`events.rs`) and the new 1s polling loop in `runtime.rs` push `set_position` / `set_playback_status` to SMTC during native playback, so they overlap. The macOS loop only refreshes metadata and lets events own position/play-state. Consider doing the same here to cut redundant COM calls.
- In `handle_windows_media_events`, `app.lock()` is held across the `player.play()/pause()/seek()` calls. Should be fine if those are non-blocking, but worth confirming they can't block on the spirc channel under contention.
- Adding `windows-media` to `default` means every Windows user pulls `smtc-tokio` (new, v0.1.0, single-maintainer). Totally fine, just flagging since it's now on the default path.

One note for visibility: CI runs `--no-default-features --features telemetry`, so none of this Windows path is compiled there. The feature-combo issue above wouldn't be caught by CI.

rlpratyoosh:
Thanks for the review!

1. I've successfully fixed the  `windows-media` feature dependency in `Cargo.toml`
2. I kept the position and state sync in the `runtime.rs` loop, but gated it behind an `if app.native_track_info.is_none()` check. If we strip it out entirely, the timeline and play state will freeze when controlling external devices, since `PlayerEvent`s stop firing.
3. You were right about the `app.lock()` in `handle_windows_media_events`, The `player.play()/pause()/seek()` calls are non-blocking; however, it's still better to not rely on them. So I scoped the `app_lock` tightly.

LargeModGames (approved):
This looks good now thanks, could you also update changelog.md?

PR #293, opened 2026-06-09
fix spotatui native device and selector behavior

### LINKED ISSUE #292: Native spotatui device missing from selector; playback fails with NO_ACTIVE_DEVICE and selector traps input
**Describe the bug**
Trying to play any track fails with a Spotify API `404 Not Found` / `NO_ACTIVE_DEVICE` error even though native streaming is enabled. After opening the device selector with `d`, the local `spotatui` playback device is missing. The device selector can also become stuck: `Enter` does not select anything and `Esc` does not return to the previous screen, so the only way out is killing and restarting the app.

  **To Reproduce**
  1. Start `spotatui`
  2. Try to play any track from the library/search/playlist view.
  3. Observe the error screen:
     `Spotify API 404 Not Found ... NO_ACTIVE_DEVICE`
  4. Press `Esc` to return
  5. Press `d` to open the device selector
  6. Observe that the `spotatui` device is missing
  7. Try pressing `Enter` on an available device entry or `Esc` to leave the selector
  8. The selector does not respond as expected, the app has to be killed and restarted.

  **Expected behavior**
Playback should start on the local native `spotatui` device when native streaming is available, or the device selector should show the local `spotatui` device so it can be selected manually.

The device selector should also remain navigable:
  - `Enter` should select the highlighted playback device or show a useful status message.
  - `Esc` should close the device selector and return to the previous screen.

   - OS: macOS
   - Terminal: kitty v0.45.0
   - Version: v0.38.6

  **Additional context**
The issue appears when Spotify reports no active playback device. The local native streaming player can be connected locally, but the Spotify devices API may not list the `spotatui` Connect device yet. In that state playback falls through to the Spotify Web API and fails with `NO_ACTIVE_DEVICE`.

shilicioo:
# Summary

Fixes #292 

This change makes explicit playback use the connected native `spotatui` streaming player when there is no current Spotify playback context, instead of falling through to the Web API and failing with `NO_ACTIVE_DEVICE`.

  It also fixes the device selector behavior:
  - Adds the live local `spotatui` device to the selector when Spotify's devices API omits it.
  - Clears stale device state on refresh.
  - Shows useful status messages for invalid/empty selection state.
  - Makes `Enter` dispatch device transfer and close the selector.
  - Makes `Esc` close the selector instead of trapping the user.

  # Testing

  - `cargo fmt --all`
    - Passed
    - Passed
    - Passed: 235 tests
  - `cargo check`
    - Passed with default features, including native streaming

  # Additional notes
The original bug showed as a Spotify API `404 Not Found` / `NO_ACTIVE_DEVICE` error when trying to play tracks, while the local `spotatui` playback device was missing from the device selector. The selector could also become stuck because `Esc` was swallowed by the global handler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 ### Fixed
 
+- **Native device selection and playback startup**: Made the local `spotatui` Connect device selectable when Spotify's devices API omits it, recovered stale native streaming sessions after long idle, fixed `Esc`/`Enter` handling in the device selector, and avoided `NO_ACTIVE_DEVICE` playback failures when native streaming is connected but no Spotify playback context is active ([#292](https://github.com/LargeModGames/spotatui/issues/292)).
 - **Search box no longer traps focus on submit**: Pressing `Enter` to run a search now always moves focus to the results list, including when re-searching while already on the Search screen (previously focus stayed stuck in the input box) ([#191](https://github.com/LargeModGames/spotatui/issues/191)).
 - **Cover-art load failures are non-fatal**: A failed album-image fetch is now logged and ignored instead of surfacing a blocking error and aborting the now-playing update, so playback metadata keeps updating when artwork can't be loaded ([#142](https://github.com/LargeModGames/spotatui/issues/142)).
 
--- a/src/core/app.rs
+++ b/src/core/app.rs
@@ -935,6 +935,10 @@ pub struct App {
   /// Reference to the native streaming player for direct control (bypasses event channel)
   #[cfg(feature = "streaming")]
   pub streaming_player: Option<Arc<crate::infra::player::StreamingPlayer>>,
+  /// Sender used to recover native streaming when a stale/disconnected player is detected.
+  #[cfg(feature = "streaming")]
+  pub streaming_recovery_tx:
+    Option<tokio::sync::mpsc::UnboundedSender<crate::infra::player::StreamingRecoveryRequest>>,
   /// Reference to MPRIS manager for emitting Seeked signals after native seeks
   #[cfg(all(feature = "mpris", target_os = "linux"))]
   pub mpris_manager: Option<Arc<crate::infra::mpris::MprisManager>>,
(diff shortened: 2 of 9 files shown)


LargeModGames:
Nice fix, this addresses the root cause well. The native-activation gate (`is_connected()` plus no `current_playback_context`, matching the saved device by both id and name) is a sensible guard, and preserving `selected_device_index` across refresh plus surfacing the API error in `get_devices` are real improvements. Good to see tests for both the Esc and Enter paths.

Two small things:

- The `Key::Esc => app.pop_navigation_stack()` arm in `select_device.rs` is now dead code. `Esc` is intercepted globally in `handle_app` then `handle_escape`, and never reaches the per-screen handler. The actual fix that makes Esc work is the new `SelectDevice` arm in `handle_escape` (which is also what `escape_exits_device_selector` exercises). Worth removing the unreachable branch or leaving a comment that global escape owns it.
- Can you confirm the non-native transfer branch below the native one also persists the device id when `persist_device_id` is set? Just want the two branches symmetric.

Otherwise looks good to merge.

shilicioo:
Thanks @LargeModGames, good catch. I removed the unreachable `Key::Esc` arm from `select_device.rs`

the non-native transfer branch is already symmetric: when the Web API transfer succeeds, it persists the selected `device_id` via `client_config.set_device_id(device_id)` whenever `persist_device_id` is true.

shilicioo:
Found one more related long-idle case and pushed a fix for it.

After leaving `spotatui` running for a long time, the native librespot/Connect session can become disconnected while `App` still holds the stale `streaming_player`. In that state playback falls through to the Spotify Web API and hits `NO_ACTIVE_DEVICE`, and the device selector cannot synthesize the local `spotatui` device because `player.is_connected()` is false.

The new change treats stored native players as active only when connected, clears stale native state when detected, and triggers the existing native streaming recovery flow from playback/device selection instead of falling through to Web API playback.

LargeModGames (approved):

PR #315, opened 2026-06-21
Fix silent native streaming (Spotify CDN HTTP 530)

LargeModGames:
## Problem
Native streaming showed tracks as "playing" but produced no audio: librespot 0.8.0 logged `Unable to load encrypted file: FailedPrecondition, StatusCode(530)` and skipped every track.

## Root cause
Spotify `storage-resolve` returns several CDN URLs per track. The first can return HTTP 530 (Cloudflare "origin unreachable") while the others serve valid audio. librespot 0.8.0 only falls back to the next URL on connection errors, not on a non-206 HTTP status, so it aborted on the first 530. Upstream issue #1725 / PR #1722 (open, unmerged, unreleased).

## Fix
- Point librespot at our maintained fork `LargeModGames/spotatui-librespot` (branch `spotatui`) via `[patch.crates-io]`. It backports PR #1722 onto the v0.8.0 tag: the 206 check moves inside the CDN-URL retry loop, so a non-206 (530) falls through to the next URL. `Cargo.lock` pins the exact commit.
- Halt native playback after repeated consecutive Unavailable tracks instead of letting librespot stampede the queue (which can rate-limit the account); reword the now-stale status message.
- CHANGELOG + README updated.

## Verification
Built with default features; confirmed audible playback on a Premium account. Logs show `Fetching <url> returned 530 ... trying next` then a successful fetch, with no `Unable to load encrypted file`.

## Notes
- The slim CI build (`--no-default-features --features telemetry`) does not compile librespot, so the patch is unused there (harmless warning).
- `cargo install` from crates.io still gets unpatched librespot (crates.io ignores `[patch]`); GitHub-release/CI/source builds get the fix. Drop the patch and bump the version once upstream ships a fixed librespot release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- **Native streaming silent playback (HTTP 530)**: Native audio no longer fails silently when Spotify's CDN returns HTTP 530 for the first of the candidate URLs it hands out. spotatui now builds librespot from a maintained fork ([spotatui-librespot](https://github.com/LargeModGames/spotatui-librespot)) that backports upstream [PR #1722](https://github.com/librespot-org/librespot/pull/1722) ([#1725](https://github.com/librespot-org/librespot/issues/1725)), falling back to the next CDN URL on any non-206 response. Repeated unavailable tracks now also halt playback instead of stampeding the queue and risking a rate-limit.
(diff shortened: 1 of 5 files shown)


Tanish431:
Just wanted to say I was going to report same thing... this spotify CDN error is happening to all the songs (I am suffering from it currently) probably account related (and librespot) error... still thanks for the rate-limitation.

LargeModGames:
I also had it happen to all songs, its librespot related. there currently is a PR open in the librespot repo but idk when it is going to get merged so i just forked librespot with the fix applied. this PR should fix the songs not playing

Issue #150, opened 2026-03-05
playerctl/dbus mpris no longer getting registered


kalebris:
**Describe the bug**
Previously I was able to control spotatui with playerctl on linux and it showed up in `playerctl -l` as a player that is currently playing.

I am building it with the following nix file:
```
{
  lib,
  rustPlatform,
  fetchFromGitHub,
  pkg-config,
  openssl,
  alsa-lib,
  dbus,
  pipewire,
  pkgs,
}:
rustPlatform.buildRustPackage rec {
  pname = "spotatui";
  version = "0.37.2";

  src = fetchFromGitHub {
    owner = "LargeModGames";
    repo = "spotatui";
    rev = "v${version}";
    sha256 = "sha256-IXNTNMTm29OEea360lWKHaeZ0J9mYwb0RQH8XMx5lRg=";
  };

  cargoHash = "sha256-er0c5Jv3hu7579G0NJubTBN1AYcZUSvct2ru75ltSZo=";

  nativeBuildInputs = [
    pkg-config
  ];

  buildInputs = [
    openssl
    alsa-lib
    pipewire
    dbus
  ];
  postFixup = ''
    patchelf \
      --set-rpath "${pkgs.lib.makeLibraryPath [
      pkgs.openssl
      pkgs.alsa-lib
      pkgs.dbus
      pkgs.pipewire
    ]}" \
      $out/bin/spotatui
  '';

  meta = with lib; {
    description = "Terminal Spotify client with playback";
    homepage = "https://github.com/LargeModGames/spotatui";
    license = licenses.mit;
    platforms = platforms.linux;
  };
}
```
I have also tried to look for dbus registrations but it also seems that it doesn't show up there.

**To Reproduce**
1. start spotatui
2. start a playback
3. issue `playerctl -l` and note that spotatui does not show up
4. try to find the appropriate dbus interface in python with which would also be empty:
```
λ python3
Python 3.13.12 (main, Feb  3 2026, 17:53:27) [GCC 15.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import dbus
>>> for service in dbus.SessionBus().list_names():
...     if 'spotatui' in service.lower():
...         print(service)
...
>>>
```
**Expected behavior**
I would expect spotatui to show up there.

**Desktop (please complete the following information):**
 - OS: Linux nixos unstable
 - Terminal foot
 - Version 0.37.2. Based on testing it broke between 0.36.0 and 0.36.1
 
[...]

PR #328, opened 2026-07-02
fix(mpris): implement volume Set and sync Volume property to D-Bus

### LINKED ISSUE #327: MPRIS: volume cannot be set via playerctl (connect_set_volume was never implemented)
**Describe the bug**

`playerctl` can read spotatui's volume over MPRIS, but setting it silently fails, or only ever oscillates within a fixed band instead of accumulating.

**To Reproduce**

1. Run spotatui with native streaming (default build, `mpris` feature enabled) and start playback.
2. Run `playerctl --player=spotatui volume` — reads the current volume correctly.
3. Run `playerctl --player=spotatui volume 0.05+` — volume does not change at all, or only ever bounces within a fixed +/-5% band instead of increasing further on repeated calls.

**Expected behavior**

`playerctl volume <value>` and `playerctl volume <delta>+/-` should set spotatui's playback volume, and `playerctl volume` should immediately reflect the new value afterward — the same way `play-pause`, `next`, `shuffle`, and `loop-status` already work correctly over MPRIS.

**Screenshots**

N/A — this is a D-Bus/MPRIS protocol issue, no visual component.

- OS: Linux
- Terminal: N/A (reproducible via `playerctl` CLI against any terminal)
- Version: 0.39.1

**Additional context**

I checked the full git history of `src/infra/mpris.rs` (only two commits have ever touched it: the initial implementation and a later "decouple from streaming and add full state sync" pass). In both, only the outgoing direction was implemented — `MprisManager::set_volume()` / `MprisCommand::Volume` push spotatui's current volume out to D-Bus so reads work. The incoming direction was never wired up: `player.connect_set_volume(...)` is never called, unlike `connect_set_shuffle` and `connect_set_loop_status`, which are. So there's no code path from an MPRIS `Set` request on the `Volume` property back into spotatui.

Two things are missing:

1. No `connect_set_volume` handler. `mpris_server::Player`'s default `set_volume` trait method just iterates its empty `set_volume_cbs` list and returns `Ok(())` — the D-Bus call succeeds silently but nothing happens. `MprisEvent` also has no `SetVolume` variant, so there'd be nowhere to route it even if the callback existed.

2. Volume changes never sync back to the `Volume` D-Bus property. Even once an incoming handler is added and routed through the native-streaming volume path (keyboard shortcuts, Lua's `spotatui.set_volume()`), the confirmed new volume never gets pushed back to MPRIS. The only code that does this, the `PlayerEvent::VolumeChanged` handler in `src/infra/player/events.rs`, depends on librespot's `Player::emit_volume_changed_event()`, which spotatui never calls. The periodic MPRIS state sync (`update_mpris_state` in `src/tui/runner.rs`) also only syncs metadata/playback-status/shuffle/loop-status, never volume. So `playerctl`'s `Get()` keeps returning a stale cached value, breaking relative adjustments like `playerctl volume 0.05+`.

Not related to #150 (that issue was about the MPRIS D-Bus service failing to register at all, after the `mpris` feature was accidentally dropped from defaults; it's closed and unrelated). Here spotatui registers on D-Bus fine, reads work, and play/pause/next/previous/shuffle/loop all work — only volume writes are affected, and always have been since MPRIS support was first added.

Fix requires two changes:
1. Register `player.connect_set_volume(...)` in `MprisManager::new()`, add an `MprisEvent::SetVolume(u8)` variant, and route it through `App::set_volume_percent` (the same entry point Lua's `spotatui.set_volume()` uses).
2. Push the confirmed volume back to the MPRIS `Volume` property immediately after every native-streaming volume change in `App::set_volume_percent` / `increase_volume` / `decrease_volume`, mirroring the existing `mpris.set_shuffle(...)` / `mpris.set_loop_status(...)` calls already present in those code paths.

Ritze03:
# Summary

Fixes #327. `playerctl` (and other MPRIS clients) could read spotatui's volume but not set it:

- `player.connect_set_volume(...)` was never registered in `src/infra/mpris.rs`, unlike `connect_set_shuffle`/`connect_set_loop_status` which are — so incoming D-Bus `Set` requests on the `Volume` property had nowhere to go and were silently dropped (`mpris_server`'s default handler just iterates an empty callback list and returns `Ok(())`).
- Even once wired up, volume changes made through the native streaming fast path (keyboard, Lua's `spotatui.set_volume()`) never synced back to the MPRIS `Volume` property. The only code that did this (`PlayerEvent::VolumeChanged` in `src/infra/player/events.rs`) depends on librespot's `Player::emit_volume_changed_event()`, which spotatui never calls. So `playerctl volume 0.05+` always computed its delta from the same stale cached value, and volume only ever bounced within a fixed +/-5% band instead of accumulating.

Changes:
- `src/infra/mpris.rs`: added `MprisEvent::SetVolume(u8)`, registered `player.connect_set_volume(...)` (converts the D-Bus `0.0-1.0` float to a `0-100` percent).
- `src/runtime.rs`: routes `MprisEvent::SetVolume` through `App::set_volume_percent(...)`, the same entry point Lua's `spotatui.set_volume()` already uses.
- `src/core/app.rs`: pushes `mpris.set_volume(...)` back to D-Bus in the native-streaming fast-path branches of `set_volume_percent`, `increase_volume`, and `decrease_volume`, mirroring the existing `mpris.set_shuffle(...)`/`mpris.set_loop_status(...)` calls already present there.
- `CHANGELOG.md`: added an entry under `Unreleased -> Fixed`.

Not related to #150 (that was the MPRIS D-Bus service failing to register at all; closed, unrelated). Here the service registers fine, reads/play/pause/next/previous/shuffle/loop all work — only volume writes were affected, and always have been since MPRIS support was first added (confirmed via the full git history of `mpris.rs`).

# Testing

Ran the exact commands from `.github/workflows/ci.yml` (default features, `--locked`), plus the slim build from `CONTRIBUTING.md`:

- `cargo check --locked`
- `cargo test --locked` (378 passed)
- `cargo fmt --all -- --check`
- `cargo clippy --locked -- -D warnings`
- `cargo test --no-default-features --features telemetry` (252 passed)
- `cargo build --features "mpris,streaming,alsa-backend"`

All clean, no warnings.

# Additional notes

I was not able to exercise this end-to-end with a live `playerctl`/D-Bus session in my environment (no session bus / audio device / Spotify session available), so this is verified by build/lint/test only, not a live manual test. Would appreciate a sanity check from someone who can run `playerctl --player=spotatui volume 0.05+` against a real session before merging.

No existing test coverage exists for `mpris.rs` or the volume functions in `app.rs` (hard to unit test without a live session bus / mixer), so this PR doesn't add new tests — happy to add coverage if there's a preferred approach for testing MPRIS interactions in this codebase.

### DIFF
--- a/src/core/app.rs
+++ b/src/core/app.rs
@@ -2080,6 +2080,14 @@ impl App {
           self.user_config.behavior.volume_percent = next_volume;
           let _ = self.user_config.save_config();
           self.pending_volume = Some(next_volume);
+
+          // Notify MPRIS clients of the change (VolumeChanged is never emitted by
+          // librespot for local mixer changes, so this is the only way the
+          // Volume D-Bus property stays in sync)
+          #[cfg(all(feature = "mpris", target_os = "linux"))]
+          if let Some(ref mpris) = self.mpris_manager {
+            mpris.set_volume(next_volume);
+          }
           return;
         }
       }
@@ -2118,6 +2126,14 @@ impl App {
           self.user_config.behavior.volume_percent = next_volume;
           let _ = self.user_config.save_config();
           self.pending_volume = Some(next_volume);
+
+          // Notify MPRIS clients of the change (VolumeChanged is never emitted by
+          // librespot for local mixer changes, so this is the only way the
+          // Volume D-Bus property stays in sync)
+          #[cfg(all(feature = "mpris", target_os = "linux"))]
+          if let Some(ref mpris) = self.mpris_manager {
+            mpris.set_volume(next_volume);
+          }
           return;
         }
       }
@@ -2161,6 +2177,14 @@ impl App {
           self.user_config.behavior.volume_percent = next_volume_u8;
           let _ = self.user_config.save_config();
           self.pending_volume = Some(next_volume_u8);
+
+          // Notify MPRIS clients of the change (VolumeChanged is never emitted by
+          // librespot for local mixer changes, so this is the only way the
+          // Volume D-Bus property stays in sync)
+          #[cfg(all(feature = "mpris", target_os = "linux"))]
+          if let Some(ref mpris) = self.mpris_manager {
+            mpris.set_volume(next_volume_u8);
+          }
           return;
         }
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - **Native streaming silent playback (HTTP 530)**: Native audio no longer fails silently when Spotify's CDN returns HTTP 530 for the first of the candidate URLs it hands out. spotatui now builds librespot from a maintained fork ([spotatui-librespot](https://github.com/LargeModGames/spotatui-librespot)) that backports upstream [PR #1722](https://github.com/librespot-org/librespot/pull/1722) ([#1725](https://github.com/librespot-org/librespot/issues/1725)), falling back to the next CDN URL on any non-206 response. Repeated unavailable tracks now also halt playback instead of stampeding the queue and risking a rate-limit.
 - **Native streaming startup device recovery**: Startup now trusts the local `spotatui` Connect device id, preserves saved external devices, activates the native device when Spotify reports no active playback, and renders an actionable idle playbar so paused native sessions restore without manually opening the device selector ([#301](https://github.com/LargeModGames/spotatui/issues/301)).
+- **MPRIS volume control**: `playerctl volume` (and other MPRIS clients) can now actually set spotatui's volume — previously only reads worked, since no `connect_set_volume` handler was ever registered. Volume changes made via the native streaming fast path (keyboard, Lua) now also sync back to the MPRIS `Volume` property immediately, so relative adjustments like `playerctl volume 0.05+` no longer compute from a stale cached value ([#327](https://github.com/LargeModGames/spotatui/issues/327)).
 
 ### Added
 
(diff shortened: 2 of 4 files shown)


src/core/app.rs
```diff
@@ -2080,6 +2080,14 @@ impl App {
           self.user_config.behavior.volume_percent = next_volume;
           let _ = self.user_config.save_config();
           self.pending_volume = Some(next_volume);
+
+          // Notify MPRIS clients of the change (VolumeChanged is never emitted by
+          // librespot for local mixer changes, so this is the only way the
+          // Volume D-Bus property stays in sync)
+          #[cfg(all(feature = "mpris", target_os = "linux"))]
+          if let Some(ref mpris) = self.mpris_manager {
+            mpris.set_volume(next_volume);
```
LargeModGames, line 2089: Nice fix, and mirroring the existing `set_shuffle`/`set_loop_status` pattern is the right call. One gap worth closing for completeness: there's a **fourth** native-streaming volume path that this PR doesn't cover.

In `change_volume` (`src/infra/network/playback.rs`, ~L1421-1435), the `IoEvent::ChangeVolume` handler has its own native-streaming branch that calls `player.set_volume(volume)` and updates `ctx.device.volume_percent` but never calls `mpris.set_volume(...)` — so it leaves the MPRIS `Volume` property stale for exactly the reason this PR is fixing here.

It's reached when these `app.rs` volume methods see native streaming as *inactive* and dispatch to the API path, but native has become active by the time `change_volume` re-checks (e.g. a keyboard volume press racing native-device activation at startup). Rare, but adding the same guarded `mpris.set_volume(volume)` in that branch would make the sync complete across all four native paths.

LargeModGames:
@all-contributors please add @Ritze03 for code, bug

PR #333, opened 2026-07-04
fix: don't panic on F13–F24 function keys

### LINKED ISSUE #332: F13–F24 function keys panic the app (key.rs from_f)
**Describe the bug**

Pressing any function key above F12 (F13–F24) crashes spotatui with a panic.

I use F13–F24 as global hotkeys in my setup, and when spotatui happens to be the focused window, hitting one of them by accident takes down the whole app instead of being ignored.

The panic comes from `Key::from_f` in `src/tui/event/key.rs`, which only maps F0–F12 and panics on anything higher:

```rust
// src/tui/event/key.rs
_ => panic!("unknown function key: F{}", n),
```

crossterm emits `KeyCode::F(n)` for `n` up to 24, and the `From<KeyEvent>` impl forwards every function key into `from_f`, so F13–F24 hit the panic arm.

**To Reproduce**
1. Launch spotatui and keep it focused.
2. Press F13 (or any function key F13–F24) — e.g. via a hotkey bound to one of those keys.
3. See the app panic:

```
thread '<unnamed>' panicked at src/tui/event/key.rs:93:12:
unknown function key: F24
```

**Expected behavior**

Function keys that aren't bound to anything should be ignored, the same as any other unhandled key — not crash the application.

**Screenshots**
N/A

 - OS: Linux
 - Terminal: kitty + zsh
 - Version: 0.40.0

**Additional context**

Fixed locally by mapping the unmapped range to the existing `Key::Unknown` catch-all instead of panicking (spotatui binds nothing to F13–F24). Happy to open a PR.

Ritze03:
## Summary

Pressing a function key above F12 (F13–F24) crashed the whole TUI. `Key::from_f` only mapped F0–F12 and panicked on anything higher, but crossterm emits `KeyCode::F(n)` up to `n = 24`. Every function key is routed through `from_f`, so an unmapped one — e.g. F24 triggered by a global hotkey while spotatui is focused — took the app down.

This maps the unmapped range to the existing `Key::Unknown` catch-all so unhandled function keys are ignored like any other unbound key. spotatui binds nothing to F13–F24, so no new enum variants are needed.

## Changes

- `src/tui/event/key.rs`: `from_f` returns `Key::Unknown` for `n > 12` instead of `panic!`; dropped the now-inaccurate `# Panics` doc line.

## Testing

- `cargo fmt --all`
- `cargo test --no-default-features --features telemetry` (326 passed)

Fixes #332

### DIFF
--- a/src/tui/event/key.rs
+++ b/src/tui/event/key.rs
@@ -70,11 +70,7 @@ pub enum Key {
 impl Key {
   /// Returns the function key corresponding to the given number
   ///
-  /// 1 -> F1, etc...
-  ///
-  /// # Panics
-  ///
-  /// If `n == 0 || n > 12`
+  /// 1 -> F1, etc... Unmapped keys (F13+) return `Unknown`.
   pub fn from_f(n: u8) -> Key {
     match n {
       0 => Key::F0,
@@ -90,7 +86,7 @@ impl Key {
       10 => Key::F10,
       11 => Key::F11,
       12 => Key::F12,
-      _ => panic!("unknown function key: F{}", n),
+      _ => Key::Unknown,
     }
   }
 }


LargeModGames:
this will be added in v0.40.2 as i just started the CD workflow for v0.40.1, sorry about that

PR #380, opened 2026-07-17
feat: client-side Spotify-style shuffle for native streaming

LargeModGames:
## What

Client-side ("Spotify-style") shuffle for native streaming. When shuffle is on and you start a **playlist, album, or Liked Songs** with spotatui as the playback device, the app now owns the shuffled play order itself (a flat, pre-shuffled URI list loaded via `LoadRequest::from_tracks` with Spirc-side shuffle off) instead of delegating shuffle to Spirc/Spotify.

## Why

Spirc/Spotify regenerate the shuffle order on every context reload. Because the native queue suspends and resumes the context on each queued track, a queue interruption could reshuffle the remaining order and **replay tracks you had already heard**. Owning the order in-app means a suspend/resume reloads the *same* order at the right index, so every track plays exactly once per pass.

## Behavior

- Shuffles the tracklist once with the selected song pinned first; everything else follows exactly once per pass.
- The full playlist / Liked Songs context loads in the **background** (capped at 3000 tracks, with a status message when truncated) so playback starts instantly from the tracks already on hand, then swaps in the complete order with a single position-preserving reload.
- Turning shuffle off restores the context's natural order and keeps your place; each **repeat-all lap** reshuffles for a fresh order, like the Spotify client.
- Manual Next/Previous, repeat-one replays, and duplicate track ids are resolved to the correct occurrence. Albums load fully up front; contexts the session can't cover (artist/show, raw search/recommendation lists) fall back to the previous Spirc shuffle.
- Slim (`--no-default-features --features telemetry`) CI build is unaffected: the pure order/merge helpers are unconditional and unit-tested, the streaming orchestration is `#[cfg(feature = "streaming")]`.

## Commits

- `feat`: the client-side shuffle session, background context fetch, queue suspend/resume integration, and lap reshuffle.
- `refactor`: post-review cleanup (dropped a write-only session field, extracted the duplicated re-anchor logic and the playlist/saved-track pagination loops).
- `docs`: CHANGELOG entry.

## Testing

- `cargo fmt --all`
- `cargo clippy -- -D warnings` (streaming) and `cargo clippy --no-default-features --features telemetry -- -D warnings` (slim)
- `cargo test --no-default-features --features telemetry` (462 passed, including the new shuffle helper tests)
- Manual smoke test of native shuffle + queue interaction still pending.

## Summary by CodeRabbit

* **New Features**
  * Added Spotify-style shuffle for native streaming playback with stable, app-owned play order.
  * Shuffle supports playlists, albums, and Liked Songs, including background loading for large collections.
* **Bug Fixes**
  * Shuffle now resumes correctly after queue suspension, repeat-all reshuffle, and repeat-one behavior (including duplicate tracks).
  * Shuffle playback previews are preserved appropriately during resume, and shuffled state is cleared when switching devices or starting a new target.


LargeModGames:
@coderabbitai All four findings are addressed in 0489994.

**1. Bind the suspended shuffle resume to the shuffle session (Major).**
`SuspendedContext::SpotifyShuffled` now carries the session `generation` alongside `resume_index`. Both suspend sites stamp `session.generation`, it is threaded through `IoEvent::ResumeNativeShuffleSession(Option<usize>, u64)`, and `resume_native_shuffle_session` applies the index only while the live session's generation still matches. A session replaced mid-drain is left running untouched; the `None` (exhausted) path clears only the session it actually suspended. The full-context-fetch remap already guards on generation before touching the suspend state.

**2. Pass `ManualSkip` from the manual Next path.**
`suspend_native_spotify_context_for_queue` now takes a `SuspendCause`. `next_track` passes `ManualSkip` (advances past the current track under repeat-one) and `handle_native_spotify_track_end` passes `AutoAdvance` (replays under repeat-one).

**3. Respect an exhausted resume index in the preview.**
The `SpotifyShuffled` arm in `queue_preview_tracks` now returns `spotify_mirror(max)` for `Some(_)` and `Vec::new()` for `None`.

**4. Regression coverage.**
Added streaming-gated tests: generation binding plus the repeat-one cause distinction (`core::app`), and both resume outcomes forwarding index and generation (`infra::queue::dispatch`).

Verified with `cargo fmt`, `cargo clippy --features streaming -D warnings`, and the full `cargo test --features streaming` suite (683 passing), plus the slim telemetry gate.

PR #389, opened 2026-07-19
fix: Crashes when audio backend fails instead of failing softly

### LINKED ISSUE #384: Crashes when audio backend fails instead of failing softly
**Describe the bug**

I get the following errors because ALSA is probably not configured correctly and didn't respond to spotatui.

```
[2026-07-18][14:10:14][librespot_core::spclient][INFO] Resolved "gew1-spclient.spotify.com:443" as spclient access point
[2026-07-18][14:10:15][librespot_connect::spirc][INFO] active device is <> with session <2cd6117fc5ff4a30952f19b7f3abd45d>
[2026-07-18][14:10:18][librespot_connect::state::context][WARN] couldn't load context info because: context is not available. type: Default
[2026-07-18][14:10:18][librespot_core::dealer][WARN] Websocket connection failed: WebSocket protocol error: Remote sent after having closed
[2026-07-18][14:10:18][librespot_core::dealer][WARN] Error while closing websocket: WebSocket protocol error: Sending after closing is not allowed
[2026-07-18][14:10:18][librespot_playback::player][INFO] Loading <Combustion> with Spotify URI <spotify:track:1zAOJRt6BryTA9Cv9HUiIz>
[2026-07-18][14:10:18][librespot_connect::spirc][WARN] SpircCommand::Activate will be ignored while already active
[2026-07-18][14:10:18][librespot_core::dealer][WARN] Wasn't able to reply to dealer request: channel closed
[2026-07-18][14:10:18][symphonia_format_ogg::demuxer][INFO] starting new physical stream
[2026-07-18][14:10:18][symphonia_format_ogg::demuxer][INFO] selected vorbis mapper for stream with serial=0x0
[2026-07-18][14:10:18][librespot_connect::spirc][INFO] active device is <6d7eedd5-d64a-4723-8a9d-5a0dad5e61f5> with session <eeaac8f2697b40e7aaa6193a84b0963d>
[2026-07-18][14:10:18][librespot_connect::spirc][ERROR] Client specified an invalid argument { Response status code: 400 Bad Request }
[2026-07-18][14:10:19][librespot_playback::player][INFO] <Combustion> (248653 ms) loaded
[2026-07-18][14:10:19][librespot_playback::audio_backend::alsa][INFO] Using AlsaSink with format: S16
[2026-07-18][14:10:19][spotatui::infra::player::streaming][WARN] Audio backend start error: Audio Sink Error Connection Refused: <AlsaSink> Device default May be Invalid, Busy, or Already in Use, ALSA function 'snd_pcm_open' failed with error 'No such file or directory (2)'
[2026-07-18][14:10:19][librespot_playback::player][ERROR] Audio Sink Error Connection Refused: <AlsaSink> Device default May be Invalid, Busy, or Already in Use, ALSA function 'snd_pcm_open' failed with error 'No such file or directory (2)'
[2026-07-18][14:10:19][librespot_playback::audio_backend::alsa][INFO] Using AlsaSink with format: S16
[2026-07-18][14:10:19][spotatui::infra::player::streaming][WARN] Audio backend start error: Audio Sink Error Connection Refused: <AlsaSink> Device default May be Invalid, Busy, or Already in Use, ALSA function 'snd_pcm_open' failed with error 'No such file or directory (2)'
[2026-07-18][14:10:19][librespot_playback::player][ERROR] Audio Sink Error Connection Refused: <AlsaSink> Device default May be Invalid, Busy, or Already in Use, ALSA function 'snd_pcm_open' failed with error 'No such file or directory (2)'
[2026-07-18][14:10:19][librespot_playback::player][ERROR] PlayerInternal poll: Invalid PlayerState
```

However this causes the program to crash and also make the terminal inoperable after.

**To Reproduce**

The ALSA errors probably happen because I am on Arch Linux and use pipewire instead.

**Expected behavior**

Fail softly when the audio backend returns an error and show an error message.

 - OS: Arch Linux 7.1.3-arch1-2
 - Terminal: alacritty 0.17.0
 - Version: 0.40.2

yhay81:
Fixes #384.

## Summary

Crashes when audio backend fails instead of failing softly

## Validation

- Mechanical gate: +46/-6, tests passed
- Adversarial review: approved

---
🤖 AI-authored PR, operated by @yhay81. 

## Summary by CodeRabbit

* **Bug Fixes**
  * In streaming mode, audio-backend startup and streaming errors are now captured and made available to the UI instead of being silently ignored.
  * When an audio-backend error is detected, native playback is paused to prevent repeated failure loops.
  * The UI shows **“Audio backend failed: {error}”** for a short time, then clears the message and resumes normal status handling.

### DIFF
--- a/src/infra/player/streaming.rs
+++ b/src/infra/player/streaming.rs
@@ -453,9 +453,16 @@ impl RecoveringSink {
 impl audio_backend::Sink for RecoveringSink {
   fn start(&mut self) -> audio_backend::SinkResult<()> {
     self.failed = false;
-    if let Err(err) = self.with_inner("start", |sink| sink.start()) {
-      *self.error.lock().unwrap_or_else(|e| e.into_inner()) = Some(err.to_string());
-      self.failed = true;
+    match self.with_inner("start", |sink| sink.start()) {
+      Ok(()) => {
+        // A working sink supersedes any stale error recorded before there was
+        // a player to surface it; leaving it would report a live failure later.
+        *self.error.lock().unwrap_or_else(|e| e.into_inner()) = None;
+      }
+      Err(err) => {
+        *self.error.lock().unwrap_or_else(|e| e.into_inner()) = Some(err.to_string());
+        self.failed = true;
+      }
     }
     // A sink error must not invalidate librespot's player thread. Keep consuming
     // packets silently so the TUI can surface the failure and remain usable.
@@ -1318,9 +1325,76 @@ fn new_device_id_string() -> String {
 mod tests {
   use super::{
     get_or_create_device_id, new_device_id_string, should_retry_with_fresh_credentials,
-    wait_for_oauth_callback_port, StreamingConnectionState,
+    wait_for_oauth_callback_port, RecoveringSink, StreamingConnectionState,
   };
+  use librespot_playback::{audio_backend, convert::Converter, decoder::AudioPacket};
+  use std::sync::Arc;
   use std::time::Duration;
+
+  struct StubSink {
+    start_result: fn() -> audio_backend::SinkResult<()>,
+  }
+
+  impl audio_backend::Sink for StubSink {
+    fn start(&mut self) -> audio_backend::SinkResult<()> {
+      (self.start_result)()
+    }
+
+    fn stop(&mut self) -> audio_backend::SinkResult<()> {
+      Ok(())
+    }
+
+    fn write(
+      &mut self,
+      _packet: AudioPacket,
+      _converter: &mut Converter,
+    ) -> audio_backend::SinkResult<()> {
+      Ok(())
+    }
+  }
+
+  fn failing_start() -> audio_backend::SinkResult<()> {
+    Err(audio_backend::SinkError::StateChange(
+      "stub backend failure".to_string(),
+    ))
+  }
+
+  fn stub_recovering_sink(
+    start_result: fn() -> audio_backend::SinkResult<()>,
+  ) -> (RecoveringSink, Arc<std::sync::Mutex<Option<String>>>) {
+    let error = Arc::new(std::sync::Mutex::new(None));
+    let sink = RecoveringSink::new(
+      move || Box::new(StubSink { start_result }) as Box<dyn audio_backend::Sink>,
+      Arc::clone(&error),
+    );
+    (sink, error)
+  }
+
+  #[test]
+  fn failing_backend_start_returns_ok_and_records_error() {
+    use audio_backend::Sink;
+
+    let (mut sink, error) = stub_recovering_sink(failing_start);
+
+    // The regression guard for #384: a failing backend must not propagate the
+    // error into librespot's player thread.
+    sink.start().expect("start() must fail softly");
+    assert!(sink.failed);
+    let recorded = error.lock().unwrap().take();
+    assert!(recorded.unwrap().contains("stub backend failure"));
+  }
+
+  #[test]
+  fn successful_start_clears_stale_backend_error() {
+    use audio_backend::Sink;
+
+    let (mut sink, error) = stub_recovering_sink(|| Ok(()));
+    *error.lock().unwrap() = Some("stale error from a previous sink".to_string());
+
+    sink.start().expect("start() should succeed");
+    assert!(!sink.failed);
+    assert!(error.lock().unwrap().is_none());
+  }
 
   #[test]
   fn oauth_callback_port_waits_for_previous_listener_to_release() {
--- a/src/infra/player/streaming.rs
+++ b/src/infra/player/streaming.rs
@@ -367,16 +367,20 @@ async fn run_spirc_supervisor(mut ctx: SpircSupervisorContext) {
 struct RecoveringSink {
   inner: Option<Box<dyn audio_backend::Sink>>,
   make_sink: Box<dyn Fn() -> Box<dyn audio_backend::Sink>>,
+  error: Arc<std::sync::Mutex<Option<String>>>,
+  failed: bool,
 }
 
 impl RecoveringSink {
-  fn new<F>(make_sink: F) -> Self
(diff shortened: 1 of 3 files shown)


src/infra/player/streaming.rs
```diff
@@ -135,7 +146,14 @@ impl audio_backend::Sink for RecoveringSink {
     packet: AudioPacket,
     converter: &mut Converter,
   ) -> audio_backend::SinkResult<()> {
-    self.with_inner("write", |sink| sink.write(packet, converter))
+    if self.failed {
```
LargeModGames, line 491: This one needs to go back to returning `Err`.

`sink.write()` blocking is librespot's only backpressure. Once `failed` latches, `write()` returns instantly forever and `PlayerInternal::poll` spins as fast as the decoder produces packets: a pegged core, `PositionChanged` far faster than realtime, then `EndOfTrack` after a few seconds and spirc starts advancing through the queue.

The runner's `player.pause()` is meant to stop that, but it isn't guaranteed to land: `route_command` returns `Err` in `SpircCommandMode::Failed`/`Shutdown` and `pause()` only `warn!`s. In that state the UI shows Paused while librespot silently races through the whole queue.

Worth noting this half isn't needed for the crash fix. Write errors take librespot's `handle_packet` error arm, which calls `handle_pause()` from a point where the `Playing` match already succeeded, so it can't reach `exit(1)`. They already failed soft.

Suggest: record the message as you do now, but still return the `Err`. Crash fixed, backpressure intact, librespot self-pauses immediately.

src/tui/runner.rs
```diff
@@ -638,6 +638,15 @@ pub async fn start_ui(
     let title_update = {
       let mut app = app.lock().await;
 
+      #[cfg(feature = "streaming")]
+      if let Some(player) = app.streaming_player.clone() {
+        if let Some(error) = player.take_audio_backend_error() {
+          player.pause();
+          app.native_is_playing = Some(false);
```
LargeModGames: This is missing most of the native-pause bookkeeping. `App::toggle_playback` (src/core/app.rs:5400-5408) does five things here:

```rust
self.pending_start_playback = None;
self.native_load_watchdog = None;
self.set_native_playback_intent(false);
player.pause();
if let Some(ctx) = &mut self.current_playback_context { ctx.is_playing = false; }
self.native_is_playing = Some(false);
```

`set_native_playback_intent(false)` is the important one: without it `desired_playing` stays `true` in the recovery snapshot, and #388 (now on main) resumes from that snapshot. Reconnect resumes, sink fails, pause, repeat. The leftover `native_load_watchdog` can also fire `force_native_streaming_recovery(true)` against a backend we already gave up on.

Suggest lifting this into an `App` method that mirrors the block above, rather than poking state from the render loop.

src/infra/player/streaming.rs
```diff
@@ -116,7 +120,14 @@ impl RecoveringSink {
 
 impl audio_backend::Sink for RecoveringSink {
   fn start(&mut self) -> audio_backend::SinkResult<()> {
```
LargeModGames, line 454: Minor: a successful `start()` clears `self.failed` but leaves the shared `error` slot populated. An error stored while `app.streaming_player` is still `None` survives and gets surfaced as a live failure on the first tick with a player present, even if audio works by then. Clearing the slot here on success closes it.

Also worth a unit test: `RecoveringSink::new` takes a `Fn() -> Box<dyn Sink>`, so a failing stub sink injects with no plumbing, and asserting `start()` returns `Ok` on a failing backend is the actual regression guard for #384. No audio device needed, so it runs in all three CI legs.

LargeModGames (changes requested):
Diagnosis is spot on, and the `start()` half is exactly right: an `Err` there flips librespot's state to `Paused` two lines before a branch that assumes `Playing`, so it hits `exit(1)` from its own player thread and bypasses our terminal restore. That's #384 and nothing app-side could have caught it.

Two changes and a rebase, then this is good to go. Details inline.

Rebase: the merge base predates #388, so the diff still adds a field next to `spirc_alive`, which no longer exists on main. Mechanical, ~3 hunks of "take both", none touching `RecoveringSink`. Worth re-running CI after, since the green checks here ran against the old base.

src/infra/player/streaming.rs
yhay81: Addressed in 0c1acfb.

src/tui/runner.rs
yhay81: Addressed in c0664b0.

src/infra/player/streaming.rs
yhay81: Addressed in c0ebfc4.

src/tui/runner.rs
LargeModGames: Extracting this into an `App` method is the right shape, thanks. I've rebased the branch onto main (`58ce109`) so #388 is in your base now, since I suspect that's why the next bit was easy to miss.

`pause_native_playback()` is missing one line from the `toggle_playback` sequence:

```rust
self.set_native_playback_intent(false);
```

It goes just before the `player.pause()`. That's the load-bearing one: without it `desired_playing` stays `true` in #388's recovery snapshot, so the next reconnect resumes into the sink we just gave up on, fails, pauses, and repeats. The other four lines mostly cover cosmetic state.

Two small things while you're in there: the doc comment says it does "the full bookkeeping the pause branch of `toggle_playback` does", which won't be true until that line is added. And `toggle_playback` still has its own inline copy of the sequence rather than calling the new method, so the two can drift again; pointing it at `pause_native_playback()` would close that.

Rebase details: four conflicts, all "take both" (the `StreamingPlayer` struct and its literal gained main's connection-state fields alongside your `audio_backend_error`, `spirc_alive` is gone from main, and the test imports took both symbols). Locally `fmt`, `clippy -D warnings` and the test suite are clean on both the default and slim legs, 724 and 476 tests. CI hadn't run on your last three commits, so that was the only coverage they'd had.

yhay81: Addressed in ea4c480.

LargeModGames (approved):

Issue #297, opened 2026-06-10
No playback - device issues?


kolemcrae:
**Describe the bug**
When I select a song to play, it shows the 404 error urging me to select a device. 

I can then select the device, but can't exit (I beleive this is being fixed) 

As a note, everytime I do this, a new entry for spotatui as the device pops up, I have 5 "devices" that are all spotatui. 

When I close and reopen the app, SOMETIMES it shows the device error, sometimes it lets me "play" a song. 

The song shows that it is playing and seems to act normally, but no audio plays. 

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Linux, Windows, macOS]
 - Terminal [e.g. kitty, gnome-terminal, powershell, cmd, etc]
 - Version [e.g. 0.30.0]

**Additional context**

Skyslycer:
I'm facing a similar issue on Manjaro Linux.

<img width="1077" height="456" alt="Image" src="https://github.com/user-attachments/assets/40faeb80-8e14-4916-ad99-beb4d32726f2" />

When I go to devices, only one spotatui device shows up (I have Spotify Premium), I can't select anything and I can't exit the menu either.

<img width="1077" height="456" alt="Image" src="https://github.com/user-attachments/assets/79509711-9dad-4866-b8e9-ac7a3003a307" />

Sometimes it just decides to work no problem, this just started happening recently.

EDIT: Tried compiling through yay instead of using the bin, same issue happens.

KngTyrannosaurus:
Can confirm this also seems to be hitting a reasonably vanilla Debian 13 install.

**Workaround:** I have managed to get playback to work for the session by choosing 'spotatui' as the playback device from the official app. Selecting songs etc. works just fine afterward, however this 'fix' does not persist between sessions. Closing spotatui and starting a new session results in the 404 error in the new session.

PR #407, opened 2026-07-28
feat: Move generated app state to XDG state/cache dirs

DinoLeung:
# Summary

Move generated runtime/app state out of the config directory and into XDG state/cache locations.

This separates user-authored configuration from app-generated state. `config.yml` remains suitable for hand editing or declarative management, while runtime changes such as volume, layout, active source, radio favorites, history, and token/cache files move to the appropriate XDG state/cache locations.

- Keeps user-authored config in `$XDG_CONFIG_HOME/spotatui`.
- Stores `state.yml`, listening history, and Spotify OAuth token cache under `$XDG_STATE_HOME/spotatui`.
- Stores native streaming credentials/cache under `$XDG_CACHE_HOME/spotatui/streaming_cache`.
- Persists volume changes through runtime state and flushes pending state saves before CLI command exit.
- Documents the updated config/state/cache ownership.

# Testing

- `cargo fmt --all`
  - 504 passed
- `cargo test`
  - 764 passed
- `cargo test --features all-sources`
  - 837 passed, 15 ignored
- `cargo clippy --features all-sources -- -D warnings`
- `cargo check`

# Additional notes

Existing config-dir app data is migrated on first use when the new target path does not already exist. This includes legacy runtime fields/radio favorites from `config.yml`, listening history, `last_session.yml`, Spotify OAuth token caches, and native streaming credentials/audio cache.

If a new state/cache target already exists, the legacy file or directory is left in place instead of being merged or overwritten; users may need to move or remove legacy files manually only in that conflict case.

## New Features
* Added XDG-compliant locations for configuration, runtime state, history, cache, credentials, plugins, and streaming data.
* Runtime settings now persist separately, including volume, shuffle, layout, active source, announcements, and saved radio stations.
* Added migration support for existing configuration and state files.

## Bug Fixes
* Improved radio-station merging, deduplication, sanitization, and protection of configured stations.
* Invalid configuration values now fall back to defaults with a warning.
* Improved secure persistence for sensitive files and streaming data.

## Documentation
* Updated configuration, plugin, keybinding, theme, streaming, scripting, setup, playlist, and migration guidance, including local playlist locations and persisted runtime settings.


src/runtime.rs
```diff
@@ -1183,7 +1261,7 @@ screens more often and cost more CPU. Animation-heavy views keep their separate
   // when unauthenticated). A free-source TUI launch tries a silent token load and
   // tolerates its absence; the user can add Spotify later via in-TUI login.
   let spotify_required = matches.subcommand_name().is_some()
-    || user_config.behavior.active_source == crate::core::source::Source::Spotify;
+    || runtime_state.active_source == crate::core::source::Source::Spotify;
```
LargeModGames, line 1299: **Blocking:** this reverts every free-source user to Spotify on upgrade, and then hard-blocks startup.

`BehaviorConfigString::active_source` was deleted and there is no `deny_unknown_fields`, so an existing `config.yml` with `active_source: Local` is silently ignored. No `state.yml` exists yet, so `RuntimeState::default().active_source` is `Source::Spotify`, which makes `spotify_required` true here.

That routes into `authenticate_with_fallback` -> `ensure_auth_token(interactive: true)`, and since the token cache also moved there is no cached token, so `src/core/auth.rs:344` opens a browser and blocks on the callback server. The first-run picker cannot rescue it either: it bails when `client.yml` exists, and `apply_selections` already wrote one via `init_default_spotify_config()`.

Net effect: someone deliberately running spotatui without Spotify cannot start the app after upgrading.

src/infra/history.rs
```diff
@@ -844,14 +835,9 @@ impl ListenRecord {
...
-  let home = dirs::home_dir().ok_or_else(|| anyhow!("No $HOME directory found for history"))?;
-  Ok(
-    home
-      .join(".config")
-      .join("spotatui")
-      .join(HISTORY_SUBDIR)
-      .join(LISTENS_FILE_NAME),
-  )
+  crate::core::paths::app_state_dir()
+    .map(|dir| dir.join(HISTORY_SUBDIR).join(LISTENS_FILE_NAME))
```
LargeModGames: **Blocking:** this orphans accumulated user data with no migration.

`history/listens.jsonl`, `last_synced_at.txt`, and `last_recap_at.txt` all move from `~/.config/spotatui/history/` to `~/.local/state/spotatui/history/`. That is the data behind `spotatui history recap`, stats, and streaks. After upgrade it reads as empty and the user gets no indication where it went. The PR body flags the token cache but not this one, and losing history feels considerably worse than a re-auth.

`src/core/persisted_playback.rs:117` has the same shape: `last_session.yml` moves, so resume-last-song quietly stops working once.

Both are covered by the rename shim described in the review summary.

src/core/user_config.rs
```diff
@@ -2096,23 +2043,19 @@ impl UserConfig {
...
       set_window_title: Some(self.behavior.set_window_title),
       visualizer_style: Some(self.behavior.visualizer_style),
-      dismissed_announcements: Some(self.behavior.dismissed_announcements.clone()),
       relay_server_url: Some(self.behavior.relay_server_url.clone()),
       sync_token: self.behavior.sync_token.clone(),
       local_music_path: self.behavior.local_music_path.clone(),
       subsonic_url: self.behavior.subsonic_url.clone(),
       subsonic_username: self.behavior.subsonic_username.clone(),
       subsonic_password: self.behavior.subsonic_password.clone(),
+      ytdlp_path: self.behavior.ytdlp_path.clone(),
```
LargeModGames, line 2052: **Blocking**, and the reason findings 1 and 2 cannot be deferred to a follow-up PR.

`build_behavior` reconstructs `BehaviorConfigString` wholesale from `self.behavior`. Since `active_source`, `shuffle_enabled`, `seen_announcement_ids`, and `dismissed_announcements` were removed from that struct, the first `save_config()` after upgrade rewrites the user's `config.yml` without them, permanently.

So a one-time seed added in a later release would find nothing left to read. The migration has to land in the same release that moves the keys.

src/tui/handlers/playlist.rs
```diff
@@ -99,8 +99,18 @@ fn remove_radio_station(app: &mut App) {
     app.set_status_message("Radio station has no stream URL".to_string(), 4);
     return;
   };
+  if app.is_configured_radio_station_url(url) {
```
LargeModGames, line 103: Existing-user regression, worth handling in the same shim.

Before this PR, `F` (favorite) wrote into `behavior.radio_stations` in `config.yml`. After it, everything in `config.yml` counts as configured, so this guard fires and refuses removal with "Radio station is configured in config.yml". Every station a user saved in-app now needs a manual file edit to remove.

The ownership model itself is good, and I like that the guard exists rather than failing silently. The issue is only the day-one transition: moving pre-existing `config.yml` stations into `state.yml` during migration would avoid it entirely.

src/core/state.rs
```diff
@@ -0,0 +1,427 @@
...
+          .with_context(|| format!("malformed runtime state file: {}", path.display()))
+      }
+    }
+    Err(e) if e.kind() == std::io::ErrorKind::NotFound => Ok(PersistedRuntimeState::default()),
+    Err(e) => Err(e).with_context(|| format!("reading {}", path.display())),
+  }
+}
+
+/// Save state atomically and privately.
+pub fn save(path: &Path, state: &PersistedRuntimeState) -> Result<()> {
```
LargeModGames, line 394: Non-blocking, but worth a deliberate call: this is a blind overwrite, where `save_config` does a read-modify-write (`user_config.rs:2205` reads the existing file and merges before writing).

Two concurrent spotatui instances will now clobber each other's volume, saved stations, and seen announcements, last writer wins. The merge previously made that mostly harmless. Probably acceptable for machine-owned state, but it is a behavior change that falls out of the refactor rather than an explicit decision.

src/core/config.rs
```diff
@@ -81,21 +79,15 @@ impl ClientConfig {
...
-        }
-
-        // Create .gitignore to protect sensitive files from being committed
+    match (
+      crate::core::paths::app_config_dir(),
+      crate::core::paths::app_state_dir(),
+    ) {
+      (Some(app_config_dir), Some(app_state_dir)) => {
+        fs::create_dir_all(&app_config_dir)?;
+        fs::create_dir_all(&app_state_dir)?;
```
LargeModGames: Non-blocking: the state dir is created here at umask mode (typically 0755) and the OAuth token cache lands in it, but only `state::save` chmods it to 0700, and only on the first state write.

`write_private_file` still puts 0600 on the token file, so real exposure is limited to directory listing. Calling `ensure_private_state_dir` here instead of a bare `create_dir_all` would close it.

Related, and not a regression: the config dir still gets a `.gitignore` and the state dir gets none. The token cache was never in the old gitignore either, so this is a pre-existing gap that moved rather than a new one. Dropping the panic-log and streaming-cache entries is correct now that neither lives there.

src/core/paths.rs
```diff
@@ -0,0 +1,160 @@
+use std::{ffi::OsString, path::PathBuf};
+
+const APP_DIR: &str = "spotatui";
+const FALLBACK_CONFIG_DIR: &str = ".config";
+#[cfg(any(feature = "streaming", test))]
+const FALLBACK_CACHE_DIR: &str = ".cache";
+const FALLBACK_STATE_DIR: &str = ".local/state";
```
LargeModGames: Nit: this is one joined component with an embedded separator, so on Windows `PathBuf::join` yields `C:\Users\x\.local/state\spotatui`. Windows APIs accept it so nothing breaks, but `.join(".local").join("state")` at the call site is cleaner and matches how `FALLBACK_CONFIG_DIR` and `FALLBACK_CACHE_DIR` are each a single component.

src/infra/radio/dispatch.rs
```diff
@@ -110,29 +113,43 @@ pub async fn route_radio_event(app: &Arc<Mutex<App>>, event: &IoEvent) -> bool {
...
   if empty {
     guard.set_status_message(
-      "No radio stations configured (behavior.radio_stations); search to find some".to_string(),
+      "No radio stations configured or saved; search to add some".to_string(),
       6,
     );
   }
 }
 
+fn merged_radio_stations(
```
LargeModGames: Nit: this duplicates `state::sanitized_radio_stations`, which applies the same trim / skip-blank / dedupe-by-URL rule using a `HashSet` rather than a `Vec` linear scan. Station lists are tiny so the perf difference is irrelevant, but two implementations of one rule can drift.

Since the merge here also needs to preserve configured-then-saved ordering, one option is to call the shared sanitizer over the chained iterator instead.

LargeModGames (changes requested):
1. Free-source users get pushed into a forced Spotify OAuth flow on upgrade (inline on `src/runtime.rs`).
2. Listening history is silently orphaned, along with `last_session.yml` (inline on `src/infra/history.rs`).
3. The window for doing this migration later closes on the first config save (inline on `src/core/user_config.rs`).

All three are covered by one startup shim, run before `state::load`:

- If `<state>/state.yml` is absent, deserialize the legacy keys off the raw `config.yml` into a small `LegacyRuntimeKeys` struct and seed `PersistedRuntimeState` (`active_source`, `shuffle_enabled`, `seen_announcement_ids`, plus the app-saved `radio_stations`).
- For `.spotify_token_cache.json`, `last_session.yml`, `history/`, and `streaming_cache/`: if the new path is absent and the old one exists, `fs::rename` it, falling back to copy across filesystems.

Finding 3 is why this belongs in this PR rather than a follow-up: once `save_config()` runs once, the legacy keys are gone for good.

## Non-blocking

Four more inline: an existing-user regression on in-app saved radio stations, blind-overwrite vs the read-modify-write that `save_config` does, state-dir permissions, and a Windows path separator nit. Plus:

- Two dedupe implementations for radio stations (`state::sanitized_radio_stations` uses a `HashSet`, `radio::dispatch::merged_radio_stations` re-implements it with a `Vec` scan). Tiny lists, so this is drift risk rather than perf.
- `dismissed_announcements` has no reader on `main` either. Good moment to drop dead state rather than carry it into a new file format.
- Volume now persists from three layers (`App::{increase,decrease,set}_volume`, each decoded dispatcher, and now `Network::change_volume`). Correct, just redundant.
- No CHANGELOG entry, against a repo that keeps detailed prose entries for user-visible changes.


Thanks for taking this on, the config/state separation is overdue.

DinoLeung:
Thanks for the thorough review, and apologies for an extra long PR. I did want to keep it short and focus on the state and cache decoupling, it quickly gets out of hand.

Will be addressing them later in the week.

src/infra/player/streaming.rs
```diff
@@ -1490,10 +1490,5 @@ mod tests {
 
 /// Helper to get the default cache path for streaming
 pub fn get_default_cache_path() -> Option<PathBuf> {
-  dirs::home_dir().map(|home| {
-    home
-      .join(".config")
-      .join("spotatui")
-      .join("streaming_cache")
-  })
+  crate::core::paths::app_cache_dir().map(|dir| dir.join("streaming_cache"))
```
LargeModGames: **Blocking:** this changes the Connect device identity for every existing user.

`get_or_create_device_id` reads `device_id` out of this directory, and its own comment says why: "so every launch and every in-app recovery registers as the same device (#297)". The directory moves config to cache with no migration, so after upgrade spotatui appears in Spotify as a brand new device. The CHANGELOG discloses a possible re-auth, not this.

`migrate_legacy_path_if_unclaimed` already handles directories, and its only directory test uses a `streaming_cache` fixture. One call here wires it up.

src/core/config.rs
```diff
@@ -104,18 +96,12 @@ impl ClientConfig {
...
-
-            # Streaming credentials
-            streaming_cache/credentials.json
             ";
           fs::write(&gitignore_path, gitignore_content)?;
         }
 
         let config_file_path = &app_config_dir.join(FILE_NAME);
-        let token_cache_path = &app_config_dir.join(TOKEN_CACHE_FILE);
+        let token_cache_path = &app_state_dir.join(TOKEN_CACHE_FILE);
```
LargeModGames, line 104: **Blocking**, same shape as the streaming cache.

The token cache moves config to state with no rename, which is the re-auth the PR body flags. Since the shim now exists and handles files, one `migrate_legacy_path_if_unclaimed` call here turns the disclosed regression into a non-event.

src/core/migrations.rs
```diff
@@ -0,0 +1,920 @@
...
+  targets: &LegacyConfigCleanupTargets,
+) -> Result<bool> {
+  let mut keys = targets.runtime_state_keys.clone();
+  if targets.radio_stations {
+    keys.push("radio_stations");
+  }
+  remove_behavior_keys_from_config(path, &keys)
+}
+
+fn remove_behavior_keys_from_config(path: &Path, keys: &[&str]) -> Result<bool> {
```
LargeModGames: **Blocking:** this deletes a key the same PR documents.

`docs/configuration.md` gains an "Internet Radio" section telling users to preconfigure stations under `behavior.radio_stations`. On the first upgraded launch the shim copies that list into `state.yml`, `state_owns_radio_stations` returns true, and this function removes the key. A hand-written block is gone in the release that documents it.

The deletion exists only to stop the `is_configured_radio_station_url` guard from blocking removal of migrated favorites. Narrower fix: have the guard fire only when the URL is config-owned and *not* state-owned. Then nothing needs deleting, and the four runtime keys do not either, since `persisted_state.X.is_some()` already makes the shim idempotent.

That also drops the unprompted startup rewrite of `config.yml`, which matters given the PR's premise that the file stays declaratively manageable: `write_private_file` plus `fs::rename` replaces a home-manager symlink with a regular 0600 file.

src/core/state.rs
```diff
@@ -0,0 +1,762 @@
...
+  }
+}
+
+/// Save a state patch atomically and privately.
+///
+/// Runtime saves are intentionally read-modify-write: most save call sites only
+/// own one field, so blindly writing a whole in-memory snapshot would clobber
+/// updates made by another running instance.
+pub fn save(path: &Path, state: &PersistedRuntimeState) -> Result<()> {
+  let mut merged = load(path)?;
```
LargeModGames: Non-blocking: a malformed `state.yml` never heals.

`save` starts with `load(path)?`, so once the file is corrupt every save fails forever. Startup only logs a warning, leaves `state_path` as `None`, and then each `flush_state_save` calls `handle_error`, which means a UI error on every volume nudge. A bad `config.yml` already falls back to defaults with a warning; suggest the same here, renaming to `state.yml.bak` and starting clean.

src/core/state.rs
```diff
@@ -0,0 +1,762 @@
...
+  write_state(path, &merged)
+}
+
+fn write_state(path: &Path, state: &PersistedRuntimeState) -> Result<()> {
+  let state = sanitized_persisted_state(state);
+  let yaml = serde_yaml::to_string(&state).context("serializing runtime state")?;
+  if let Some(dir) = path.parent() {
+    crate::core::paths::ensure_private_dir(dir)?;
+  }
+  let tmp = path.with_extension("yml.tmp");
```
LargeModGames: Non-blocking, and only a partial close on the concurrency finding from last round.

Read-modify-write fixes clobber-by-omission, but two instances still lose updates: both load, both merge, both write. And the temp name is fixed, so two processes can write the same `state.yml.tmp` and rename each other's half-written file into place. A pid suffix is cheap. `remove_behavior_keys_from_config` has the same shape.

src/core/persisted_playback.rs
```diff
@@ -113,14 +114,12 @@ pub enum PersistedPlayback {
...
-/// `<config dir>/last_session.yml` next to the app config.
+/// `<state dir>/last_session.yml`.
 pub fn default_session_path() -> Result<PathBuf> {
   if let Ok(path) = std::env::var(PATH_ENV) {
     return Ok(PathBuf::from(path));
   }
-  crate::core::user_config::default_app_config_dir()
-    .map(|dir| dir.join(FILE_NAME))
-    .ok_or_else(|| anyhow!("cannot resolve the spotatui config directory"))
+  crate::core::migrations::state_file_path_with_legacy_config_rename(Path::new(FILE_NAME))
```
LargeModGames: Non-blocking: this makes a path getter do filesystem work.

`state_file_path_with_legacy_config_rename` does `create_dir_all`, an unconditional `chmod 0700`, two stats, and a possible rename, on every call. That is once per throttled session save during playback and once per listen append in `history.rs`, forever, for a one-time upgrade. Running the shims once at startup would keep the getters pure and make the module easier to delete later, which is what its own doc comment asks for.

examples/plugins/README.md
```diff
@@ -21,18 +21,28 @@ privileges, so read anything you install from elsewhere (see
 Single-file plugins go straight into `plugins/`:
 
 ```bash
-cp track-notifier.lua ~/.config/spotatui/plugins/
+case "${XDG_CONFIG_HOME:-}" in
+  /*) config_home="$XDG_CONFIG_HOME" ;;
```
LargeModGames: Nit: this is more shell than the thing it installs.

The five-line `case` block is now in both README snippets and six plugin header comments, replacing a one-line `cp`. `${XDG_CONFIG_HOME:-$HOME/.config}` covers the unset case inline; the relative-value case is rare enough to state once in `PLUGINS.md`. CodeRabbit asked for correctness here and got it, but the example headers are the first thing a plugin author reads.

LargeModGames:
Sorry this keeps growing, i know you've already done more rounds than you signed up for. If you'd rather hand any of it off, i'll push the fixes myself.

DinoLeung:
Thanks man. I’m happy to take another pass at the changes, but if there’s another round after that, I’d be happy to hand the remaining fixes over to you.

DinoLeung:
I've made another round of changes.

I'll leave https://github.com/LargeModGames/spotatui/pull/407#discussion_r3689670447 and https://github.com/LargeModGames/spotatui/pull/407#discussion_r3689670448 to you, since I'm not confident I can make a set of changes there that would be up to the project's standard.

Also, I found it funny that you read my intent here as making the project compatible with Home Manager configuration.

LargeModGames (approved):

PR #471, opened 2026-08-16
feat(tui): fuzzy filter for the settings screen

### LINKED ISSUE #469: Add a fuzzy finder for the settings menu
It takes forever to navigate the settings menu

The title is pretty self explanatory. I would like if there's a fuzzy finder to easier find the settings I'm looking for instead of having to scroll through the setting menu just to find a single setting

LargeModGames:
Closes #469.

## What

Press the search key (`/` by default) in Settings to filter the rows instead of scrolling for one.

- The query is a fuzzy subsequence: `volinc` finds **Volume Increment**, `skms` finds **Seek Duration (ms)**. Whitespace in the query is ignored; smart-case like the Help filter.
- Rows match on their name and their `config.yml` key (the shared `behavior.`/`keys.`/`theme.` prefix only counts once the query itself is dotted, otherwise `vol` would match most of the tab through `beha[v]i[o]r.`). Descriptions are deliberately not searched: they are prose the screen never draws, so a row kept on one would show no reason for being there.
- Results are ranked best-first (name match outranks id match; tight runs and word starts score higher) with the matched characters highlighted.
- `Enter` applies the filter, `Esc` clears it before it means "leave Settings", `←`/`→` keep the filter so one query can be walked across tabs. Mouse clicks and scrolls resolve against the filtered rows.
- The query renders on the second line of the Controls block while it exists.

## How

- `src/tui/fuzzy.rs`: small subsequence matcher (best placement over candidate starts, boundary/consecutive bonuses, leading penalty) returning score + merged byte ranges.
- `src/tui/handlers/filter_input.rs`: the Help menu's filter-input editing keys (`Backspace`, `Ctrl+W`, `Ctrl+U`, ...) hoisted out so both screens share them; `help_menu.rs` now calls it.
- `popups::highlighted_spans` (was `highlighted_help_line`) is shared with the settings list for highlight rendering.
- `App::open_settings_screen()` is the single front door (keybinding, sidebar click, Lua `navigate`), so no entry can carry a stale filter in.
- `settings_layout` is shared between `ui/settings.rs` and `handlers/mouse.rs` so drawn and clickable rows cannot drift.

## Checks

- `cargo fmt`, slim `clippy -D warnings`, slim `cargo test` (629 passed), default `cargo check` all green.
- `tools/gates.count`: `app_field_writes_in_tui_handlers` 786 → 785, `test_attribute_total` 1385 → 1404.

## Summary by CodeRabbit

* **New Features**
  * Added fuzzy filtering to the Settings screen by setting name or configuration key.
  * Added ranked results, highlighted matches, result counts, and a no-results message.
  * Added keyboard controls for searching, editing, applying, clearing, and canceling filters.
  * Filters persist while switching categories and reuse the configured search key.
  * Added Help-menu guidance for Settings filtering.

* **Documentation**
  * Updated configuration and keybinding documentation with filtering behavior.


LargeModGames:
Added the assertion in ca95f86: the click test now checks the query survives (`settings_filter == "sort"`), so a future `apply_settings_filter` that clears the query would fail it.

PR #498, opened 2026-08-28
fix(audio): stop two pump freezes, and enable the alternative sources on macOS

### LINKED ISSUE #496: Losing the audio output device freezes the app (all decoded sources)
**Describe the bug**

While a decoded source is playing (Local Files, Subsonic, Internet Radio, YouTube, Qobuz), losing the audio output device — Bluetooth headphones disconnecting, AirPods going back in their case, a USB DAC unplugged — does not just make playback silent. It freezes the whole app: the UI still draws and the playbar still says "playing", but search stops returning results and the playback controls do nothing. Only a restart recovers it.

The cause is that these five sources share one audio engine (`src/infra/audio/player.rs`), and rodio's `Player::clear()` and `try_seek()` wait on the audio callback with **no timeout** (`sleep_until_end()` → `recv()`, and the seek feedback channel). Once the device is gone that callback never runs again, so the wait never ends. Those calls are made straight from the serial IoEvent pump (`play_file`, `play_prepared` and `stop` all call `clear()`), so the one that blocks takes every unrelated request queued behind it down with it — which is why searching dies along with playback.

Worth noting that this has two shapes, and the one that bites is the one nothing reports: cpal notices a device being *removed* (`DeviceNotAvailable`), but it cannot notice the OS simply moving its **default output** elsewhere. In that case the device we opened is still perfectly alive and the stream keeps feeding it, so there is no error anywhere — just silence, then a freeze at the next transport call.

**To Reproduce**

1. Build with any decoded source, e.g. `cargo run --features all-sources`
2. Start playing a local file, a YouTube track, or a radio station through Bluetooth headphones
3. Disconnect the headphones mid-track (put AirPods back in their case, or unplug a USB DAC)
4. Audio goes silent while the playbar still shows "playing"
5. Try to search, or press play/pause, or skip a track — nothing responds from here on

**Expected behavior**

Losing the output device should not be able to freeze the app. Ideally playback continues on whatever output the system moved to (paused, the way macOS itself behaves when AirPods come out); at minimum the app stays responsive and says what happened.

 - OS: macOS 26.6.2 (arm64) — but see below, this is not platform-specific
 - Terminal: any (not terminal-specific)
 - Version: 0.41.0 (current `main`)

**Additional context**

Not a macOS-specific bug, even though that is where I hit it. `LocalPlayer` is shared by all five sources on all platforms, and the blocking calls are unguarded on `main` (`src/infra/audio/player.rs` lines 91, 164, 188 and 233). A USB DAC unplugged on Linux, or Bluetooth headphones disconnecting on Windows, goes down the identical path, so **the released Linux and Windows binaries have the same exposure today**. (macOS release binaries do not ship the decoded sources yet, which is the only reason this has not been reported from there.)

I have a fix and will open a PR referencing this issue.

### LINKED ISSUE #497: Some radio stations freeze the app instead of failing to play
**Describe the bug**

Some internet radio stations do not just fail to play — they freeze the app. Selecting one plays nothing and reports no error, and from that moment search and the playback controls stop responding (audio that was already playing keeps going). Only a restart recovers it.

Starting a station ends by working out the stream's audio format, and rodio's symphonia does that by scanning the stream for a start-of-stream marker it recognises. A live stream has no end to stop that scan. So a station in a format symphonia cannot identify leaves it reading forever — and since `play_stream` is awaited on the serial IoEvent pump (`src/infra/radio/dispatch.rs:276` on `main`), everything queued behind it stops too.

The specific format that triggered it here is MPEG-2 ADTS AAC, which is common on European radio. symphonia's `AdtsReader` registers only the MPEG-4 ADTS marker `ff f1`, not MPEG-2's `ff f9` (`symphonia-codec-aac/src/adts.rs`), so a stream starting with `ff f9` never matches anything. Worth knowing that a mime hint cannot rescue this: `Probe::format` takes the hint as `_hint` and ignores it.

**To Reproduce**

1. Build with radio, e.g. `cargo run --features internet-radio`
2. Search the station directory for `radio bruno`
3. Select any of the results — for example `https://stream3.xdevel.com/audio6s975355-281/stream/icecast.audio` (`Content-Type: audio/aacp`, first bytes `ff f9`)
4. Nothing plays and no error appears
5. Try to search again or change track — the app no longer responds

Measured directly: `open_radio_stream` returns in under a second (content type and ICY station name both read correctly), and `LocalPlayer::prepare_stream` then never returns. A thread sample shows it parked inside `read`, waiting on stream-download for bytes it will never be able to make sense of.

**Expected behavior**

A station that cannot be decoded should fail quickly with a message, and leave the app usable. `src/infra/radio/stream.rs` already caps the connect phase (`CONNECT_TIMEOUT`) and the header wait (`HEADER_TIMEOUT`) for exactly this reason — the format probe is the third step of the same tune-in sequence and is the one left unbounded.

 - OS: macOS 26.6.2 (arm64) — but see below, this is not platform-specific
 - Terminal: any (not terminal-specific)
 - Version: 0.41.0 (current `main`)

**Additional context**

Not platform-specific: the unbounded await is in shared code on `main`, so **the released Linux and Windows binaries freeze on these stations today**.

Two things I found while tracking it down that are worth recording:

- Giving up has to stop the **download**, not the reader. A cancellation flag the reader checks between reads is never reached, because the probe parks *inside* `read`. Cancelling the stream-download task marks the stream done and wakes every waiter, so the read returns, the probe hits end-of-stream, and the thread and its download are released together.
- Making these stations actually *play* is an upstream matter: adding `ff f9` to symphonia's ADTS markers is a one-line change there and would cover a good slice of European radio. Bounding the probe here only turns the freeze into a clean error.

I have a fix for the freeze and will open a PR referencing this issue.

alessandro-zanni:
Two freezes and a platform. The freezes are the urgent half — they affect the released Linux and Windows binaries today; macOS enablement is what uncovered them.

**Fixes #496** (losing the audio output device) — **Fixes #497** (an undecodable radio station)

### The freezes

Both are the same class: an unbounded wait on the serial IoEvent pump. When one of those never returns it takes every unrelated request queued behind it, so the app looks frozen — search dead, transport dead — rather than merely silent.

- **#496**: rodio's `clear()` and `try_seek()` wait on the audio callback with no timeout. Lose the output device and that callback never runs again. `LocalPlayer` now refuses those calls once the device is gone, and the driver's tick rebuilds the output on the new default device and restages the track there, paused where it stopped — what macOS itself does when AirPods come out.
  Detection has two shapes and only one is an error anybody reports: cpal sees a device *removed*, but it cannot see the OS moving its **default output** elsewhere, which is the common case and leaves the stream feeding a device nobody hears. So the sink also remembers what it opened and compares against the current default.
- **#497**: rodio's symphonia probes the format by scanning for a marker it recognises, and a live stream never ends. `stream.rs` already caps connect and header waits for exactly this reason; the probe was the third step of that sequence and was unbounded. Now capped — and giving up cancels the **download**, not the reader, because the probe parks *inside* `read` where a flag would never be seen.

### macOS

`LocalPlayer::open_sink()` bailed on macOS. That bail was never a fix for an observed crash in this engine: it has been there since the commit that introduced the player, inherited from #9/#20, which were librespot-playback's own `rodio-backend` on an older rodio. The lockfile still carries both — librespot pulls rodio 0.21/cpal 0.16, this player uses rodio 0.22/cpal 0.17 and the rewritten `DeviceSinkBuilder` API. cpal already ran on macOS in every shipped build via `audio-viz-cpal`, and `route_decoded_macos_event` was already written and unreachable.

Native Spotify streaming is untouched and keeps `portaudio-backend`.

**This changes what macOS release binaries contain**: `cd.yml`'s two macOS rows gain the five sources, so those binaries get bigger and the YouTube source wants `yt-dlp`, matching the Windows row.

### What is verified, and what is not

Nothing in CI covered any of this — all seven legs are `ubuntu-latest`. This PR adds a `macos-latest` check + clippy job using cd.yml's macOS feature set, since that is the only leg that compiles the `cfg(target_os = "macos")` arms, portaudio, `macos-media` and `audio-viz-cpal`. No `test` job: the suite is platform-independent logic the Linux legs already run, and the device tests are `#[ignore]`d because runners have no audio output.

So the audio path itself was verified **by hand on real hardware** (M-series Mac, macOS 26.6): local files, radio and YouTube playing over built-in output and over Bluetooth, headphones disconnected mid-track, and the previously-hanging station. Seven `#[ignore]`d live tests that drive the real sink (radio, YouTube, Qobuz) pass locally, plus six device tests in `player.rs` — including one timeout-asserted on a worker thread, because a regression there deadlocks instead of failing.

The five ignored Subsonic live tests fail on `main` too: `demo.navidrome.org` now returns an empty playlist first, so `playlists[0]` has no tracks. Unrelated, untouched.

### Deliberate limits

- Stations in formats the bundled decoder cannot identify **still will not play** — they now fail in a second or two instead of freezing. Making MPEG-2 ADTS AAC work is a one-line upstream change in symphonia (its `AdtsReader` claims only `ff f1`, not `ff f9`); worth filing there, not worth a fork here.
- On a device change, radio is torn down rather than restaged: a live stream has no position to return to.
- A track playing from the native queue resumes at the *next* queued item; there is no "replay this queue item" event and inventing one felt out of scope.
- `open_sink` no longer falls back to sweeping every other output device (that path required rodio's helper, which installs an `eprintln!` error callback that corrupts the TUI). It still falls back across the default device's other configs; a machine whose default output cannot be opened now gets a clear error instead of audio from a surprise device.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XTSifDzSFpuSUyeBTJWUii

## Summary by CodeRabbit

* **New Features**
  * macOS releases now support Local Files, Subsonic, Internet Radio, YouTube, and Qobuz.
  * macOS media keys and Now Playing support these sources without requiring a Spotify login.
  * Playback resumes its position and pause state after audio-device recovery.

* **Bug Fixes**
  * Playback no longer freezes when an audio device is disconnected.
  * Failed recovery cleans up queued playback correctly.
  * Unrecognized radio streams now fail promptly instead of hanging.

* **Documentation**
  * Updated macOS installation and platform-support guidance.


src/core/driver/mod.rs
```diff
@@ -388,6 +388,116 @@ impl Driver {
...
+            // hold the advance latch or the block below would read that as
+            // end-of-track and skip a song while the replay is in flight.
+            if let Some(s) = $app.$playback.as_mut() {
+              s.advancing = true;
+            }
+            $app.dispatch(IoEvent::ReplayCurrentTrack);
+            if resume_ms > 0 {
+              $app.dispatch(IoEvent::Seek(resume_ms));
+            }
+            $app.dispatch(IoEvent::PausePlayback);
```
LargeModGames, line 442: This pauses on every default-output change, in both directions. `device_lost()` is also true when the name moved because the user plugged headphones in, and the recovery then ends on `PausePlayback` with `playback paused here`. macOS itself only pauses on removal.

The signal to split the two cases is already here: CoreAudio and WASAPI both raise `StreamError::DeviceNotAvailable` on removal, which sets `lost`. Pause only when `lost` was set. When only the name changed, restage and keep the play/pause state the session had.

src/infra/radio/dispatch.rs
```diff
@@ -270,10 +270,31 @@ async fn start_radio(app: &Arc<Mutex<App>>, uri: &str) {
...
+  let (reader, mime, cancel) = (opened.reader, opened.content_type, opened.cancel);
   let decode_player = Arc::clone(&player);
-  let result =
-    tokio::task::spawn_blocking(move || decode_player.play_stream(reader, mime.as_deref())).await;
+  // Bound the probe, the last unbounded step of tune-in. It scans for a marker
+  // it recognises and a live stream never ends, so an unsupported codec would
+  // otherwise scan forever — and this is awaited on the serial pump, where that
+  // is not silence but a frozen app. Cancelling on the way out is what lets the
+  // probe thread and its download go, since `spawn_blocking` cannot be aborted.
+  let result = tokio::time::timeout(
```
LargeModGames, line 280: `timeout` abandons the `spawn_blocking` closure but cannot stop it, and the closure is `play_stream`, which is `prepare_stream` plus `play_prepared`. `acquire_player` reuses the current radio player, and `radio_playback` still names the old station during this probe. A probe that matches on the prefetched bytes after the 10 s mark appends the new station to the shared sink: the error shows, the playbar names the old station, the new one is audible, and no transport control reaches it.

Keep only `prepare_stream` inside the timed closure, and call `play_prepared` here after the timeout check.

src/infra/audio/player.rs
```diff
@@ -60,15 +117,120 @@ impl LocalPlayer {
...
+  /// cutting the second one short would break slow playback to fix a freeze. So
+  /// the wait re-*asks the device* every [`AUDIO_THREAD_POLL`] instead of
+  /// guessing, gives up at once when it really went away, and only past
+  /// [`AUDIO_THREAD_CEILING`] — nothing identifiably wrong, still no answer —
+  /// declares it lost anyway, because a pump that never returns is worse than
+  /// a track that never plays.
+  ///
+  /// A call that truly never returns leaves its worker parked for the life of
+  /// the process. That is the cheap half of this trade.
+  fn bounded<T: Send + 'static>(&self, op: impl FnOnce() -> T + Send + 'static) -> Option<T> {
```
LargeModGames, line 185: The poll reads `self.device_lost()` and the ceiling arm stores `lost = true` into whichever `Sink` is current, but `reopen()` replaces the whole `Sink` under the same mutex. If the tick reopens during a wait, the poll then reads the fresh sink (false), waits the full 90 s, and marks the healthy new sink lost.

Capture the old sink's `lost` flag and `Arc<Player>` at entry. On each poll, return `None` at once when the current player is no longer `Arc::ptr_eq` to the captured one. The same window exists in `play_file`, which captures the `Arc<Player>` before the decode and appends to it after.

src/core/driver/mod.rs
```diff
@@ -388,6 +388,116 @@ impl Driver {
...
+      };
+      if lost {
+        let reopened = match app.queue_now.as_ref() {
+          Some(QueueNowPlaying::Decoded(d)) => d.player.reopen().is_ok(),
+          _ => false,
+        };
+        if !reopened {
+          app.queue_now = None;
+        }
+        app.set_status_message("Audio output device disconnected.", 8);
```
LargeModGames, line 497: Two problems in this arm. On a successful reopen the fresh sink is empty, so the queue-advance block below fires and the next queued item plays out loud on the new device, under a message that says `disconnected`. That is the opposite of the paused restage the context path does. On a failed reopen, `app.queue_now = None` bypasses `resume_or_finish`, so `queue_suspended` and the remaining `native_queue` items are stranded.

Pause after the advance (or restage), move this message into the failure arm, and route the failure through the same teardown `resume_or_finish` does.

src/infra/audio/player.rs
```diff
@@ -60,15 +117,120 @@ impl LocalPlayer {
...
+
+  /// Rebuild the output on the *current* default device and drop the dead one.
+  ///
+  /// The new sink starts paused and empty, at the volume the old one had, so
+  /// the caller restages whatever was playing (see the driver's tick). Dropping
+  /// the old `Sink` releases its keepalive, which lets the old audio thread
+  /// exit; `Player`'s own `Drop` only sets a flag, so nothing here blocks.
+  ///
+  /// **Blocking:** opens a device, like [`new`](Self::new).
+  pub fn reopen(&self) -> Result<()> {
```
LargeModGames, line 223: `reopen()` runs from `Driver::tick`, under the App lock on the UI thread, and `open_sink` waits on `init_rx.recv()` with no timeout. A device open takes tens of milliseconds in practice, so this is a backstop rather than a hang I can show. A `recv_timeout` of a few seconds in `open_sink` bounds it, and also bounds `LocalPlayer::new`. That is enough here; moving the reopen off the tick is more than this PR needs.

LargeModGames (commented):
Thanks, the two freeze fixes are right in mechanism, and I verified them on Windows: clippy with the five sources is clean, the six ignored device tests pass on the real WASAPI output, and the new radio test gives up at 10.9 s against the live station. Five things in the recovery path to change before merge, inline below.

src/core/driver/mod.rs
alessandro-zanni: Split, and now it only pauses on the removal half.

`LocalPlayer::device_removed()` reads the `lost` flag on its own (the one cpal's error callback raises on `DeviceNotAvailable`); `device_lost()` is unchanged and still covers both shapes, so the tick still notices a default that moved. The recovery reads `device_removed() || is_paused()` *before* the reopen (the fresh sink is paused and reports no removal) and dispatches `PausePlayback` only then — headphones in keeps playing on the new device, with "Audio device changed - playback moved to it." instead. Same split applied to the queue slot below.

src/infra/radio/dispatch.rs
alessandro-zanni: Fixed. The timed closure is now `LocalPlayer::prepare_stream(reader, mime, None)` only — it builds a decoder and touches no sink — and `play_prepared` runs after the timeout check, in its own `spawn_blocking`. A probe that matches late now has nowhere to append.

`LocalPlayer::play_stream` was the thing that welded the two together and had no other production caller, so it is gone; `live_stream_plays_through_sink` calls the pair directly.

src/infra/audio/player.rs
alessandro-zanni: Done. `bounded()` clones the `Arc<Player>` and the `lost` flag out of the sink at entry, and on each poll returns `None` the moment the current player is no longer `Arc::ptr_eq` to that one — before the `device_lost()` check, so a reopened sink never gets waited on and the ceiling arm stores into the *captured* flag rather than whatever is current.

`play_file` re-checks the same identity after the decode and bails with "audio output device changed while decoding" instead of appending to the sink nobody plays through; the restage the tick dispatches lands on the new one.

New `#[ignore]`d test `a_wait_gives_up_when_the_sink_is_reopened_under_it` covers it: reopen while a 120s wait is parked, assert it gives up within one poll and the fresh sink is not lost. Passes here on CoreAudio along with the other six.

src/core/driver/mod.rs
alessandro-zanni: Both fixed.

Success: the arm now only records `Some(pause_after)`; the advance block below runs as before, and a settle step after it dispatches `PausePlayback` (same removal-vs-moved rule as the context path) so the next queued item does not start out loud. The pump is serial, so it lands behind the `AdvanceNativeQueue`.

Failure: new `IoEvent::FinishNativeQueue`, consumed by `route_queue_event` and calling `resume_or_finish` — the same teardown a drained queue runs, so the slot is stopped, `queue_suspended` is resumed and `native_queue` keeps its items. The "disconnected" message moved into this arm.

src/infra/audio/player.rs
alessandro-zanni: Bounded in `open_sink`: `init_rx.recv_timeout(DEVICE_OPEN_TIMEOUT)` with `DEVICE_OPEN_TIMEOUT = 5s`, so it covers `reopen()` on the tick and `LocalPlayer::new` in the same place. Left the reopen on the tick as you said.

alessandro-zanni:
Pushed `246db13` with all five, plus the three CodeRabbit nits.

| Review point | Fix |
|---|---|
| Pauses on every default-output change | `device_removed()` (cpal's `lost` flag alone) splits removal from a moved default; pause only on removal or an already-paused session |
| Probe closure keeps running and appends to the shared sink | only `prepare_stream` inside the `timeout`; `play_prepared` after the check. `play_stream` removed (it welded the two together, no other caller) |
| `bounded()` / `play_file` race with `reopen` | both capture the sink identity and bail on `!Arc::ptr_eq`; the ceiling marks the captured `lost` flag |
| Queue arm: advance plays out loud / failure strands state | pause settles after the advance dispatch; failure routes through the new `IoEvent::FinishNativeQueue` → `resume_or_finish` |
| Unbounded `init_rx.recv()` | `recv_timeout(5s)`, bounding `reopen` on the tick and `LocalPlayer::new`. Reopen stays on the tick, per your call — @coderabbitai this also answers the "keep device reopening off the tick path" comment |

CodeRabbit nits: `dtolnay/rust-toolchain@stable` in the macOS job (actions-rs is archived / node12), the stale `cd.yml` feature-matrix comment, and the radio CHANGELOG heading.

Verified: fmt, clippy `-D warnings` on slim / all-sources / headless, `cargo test` on slim (907) and all-sources (1341), and on this Mac the seven `#[ignore]`d device tests plus the five live radio ones — including the new `a_wait_gives_up_when_the_sink_is_reopened_under_it`. `tools/gates.count`: `test_attribute_total` 1758 → 1759. CHANGELOG and the three instruction files describe the new pause rule.

src/infra/network/mod.rs
```diff
@@ -143,6 +143,23 @@ pub enum IoEvent {
...
+  #[cfg_attr(
+    not(any(
+      feature = "local-files",
+      feature = "subsonic",
+      feature = "qobuz",
+      feature = "youtube"
+    )),
+    allow(dead_code)
+  )]
+  FinishNativeQueue,
```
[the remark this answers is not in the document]
alessandro-zanni: Not reachable, so leaving both out of the list.

`route_queue_event` is compiled unconditionally and runs *before* the Spotify handler in the pump (`runtime/pump.rs`), and its `AdvanceNativeQueue` / `FinishNativeQueue` arms return `true` unconditionally — no cfg, no early return between the claim gate and the router. So neither variant can reach `handle_network_event`, and the auth gate in front of it never sees them. The `=> {}` arm is the belt-and-braces the comment above it describes, matching how `AdvanceNativeQueue` has always been wired; adding them to `event_bypasses_spotify_auth` would document a path that does not exist.

LargeModGames (commented):
Thanks, `246db13` does what each of the five comments asked, and the radio and `bounded()` fixes are right. One more round on the recovery mechanism before merge. The inline comments below share one root cause: recovery expresses "stay paused" as a `PausePlayback` queued ahead of async work that ends in `sink.play()`. A desired-play-state flag on the session that the restage, resume and commit paths read, the way `queue_slot_desired_playing` already works for the Spotify slot, fixes the context arm, the queue settle step and the suspended-context resume in one place.

Two findings sit in files outside this diff, so no inline anchor:

**`bounded()` waits still run under the App lock.** The 3 s poll / 90 s ceiling is reached while the tokio App mutex is held: `commit_fetch` in `src/infra/qobuz/dispatch.rs` holds `app.lock()` across `player.play_prepared`, the queue `resume_local` across `player.seek`, `teardown_local` / `teardown_radio` across `player.stop` through if-let temporaries, and local `play_index` calls `player.stop()` on the serial pump. `tui/runner.rs` takes the same lock on every tick, key and draw, so the infinite freeze became a 90 s freeze in the case `bounded()` is documented as the backstop for. `play_prepared`'s own doc says to call it off the App lock (see `stop_detached`); the other sites need the same treatment.

**macOS media keys still need a Spotify session.** `MacMediaManager` registration in `src/runtime/startup.rs:348` is gated on `streaming_attempted`, justified by the comment "macOS plays no decoded source", while Windows registers unconditionally so decoded sources get media keys. A macOS user who picks Qobuz or Local Files in the first-run picker and skips the Spotify login (supported since #495) gets no media keys, no Now Playing and no AirPods play/pause for a source that now plays fine, and `route_decoded_macos_event` is unreachable. Same stale prose: `player.rs:124` ("or on macOS"), `core/first_run.rs:11` and `:55`.

Not blocking, a follow-up issue is fine for these:

- `from_device(..).open_sink_or_fallback()` drops rodio's `open_default_sink` sweep over the other outputs. An HDMI default on a powered-off monitor, or a WASAPI exclusive hold, now fails every decoded source and every `reopen`.
- On Linux/ALSA cpal's default device is the constant "Default Audio Device", so the name compare can never fire there and only `DeviceNotAvailable` works.
- Radio `play_prepared` returns `()` and no-ops on a lost device, and the caller publishes `radio_playback` and shows the station live anyway.
- Local `play_file` device errors go to `fail_index`, which poisons the track and ends a one-track session with "no playable tracks left".

src/core/driver/mod.rs
```diff
@@ -388,6 +388,150 @@ impl Driver {
...
+          if reopened {
+            // The fresh sink is empty, so `is_finished()` is true from here on:
+            // hold the advance latch or the block below would read that as
+            // end-of-track and skip a song while the replay is in flight.
+            if let Some(s) = $app.$playback.as_mut() {
+              s.advancing = true;
+            }
+            $app.dispatch(IoEvent::ReplayCurrentTrack);
+            if resume_ms > 0 {
+              $app.dispatch(IoEvent::Seek(resume_ms));
```
LargeModGames: `ReplayCurrentTrack` can fail on the fresh sink (a second default flap during the decode, a rejected format), and every per-source `replay_current` tears its session down on failure. The `Seek` and `PausePlayback` queued behind it then find no decoded session and no queue slot, fall through the router chain to `Network::handle_network_event`, and seek the user's real Spotify player to `resume_ms` and pause it. Neither event is in `event_bypasses_spotify_auth`, so a Spotify-free build shows two "Not connected to Spotify" toasts for an unplugged cable instead. Do not queue these blind from the tick: have the replay path apply the seek and the pause state itself once the track is staged, or dispatch them only from its success arm.

src/core/driver/mod.rs
```diff
@@ -388,6 +388,150 @@ impl Driver {
...
+            .as_ref()
+            .is_some_and(|s| s.player.reopen().is_ok());
+          if reopened {
+            // The fresh sink is empty, so `is_finished()` is true from here on:
+            // hold the advance latch or the block below would read that as
+            // end-of-track and skip a song while the replay is in flight.
+            if let Some(s) = $app.$playback.as_mut() {
+              s.advancing = true;
+            }
+            $app.dispatch(IoEvent::ReplayCurrentTrack);
```
LargeModGames: The removal half still expresses "stay paused" as a `PausePlayback` queued behind work that ends in `sink.play()`. The replay plays before the pause drains, so a removed device still gets an audible burst on the new default. Qobuz mid-download is worse: `replay_current` returns early ("still downloading"), the pause pauses the empty sink, then `commit_fetch` computes `was_paused = tempfile.is_some() && is_paused()`, which is false in that window, and `play_prepared` starts the track on the laptop speakers seconds after the user unplugged. Replace the queued pause with a desired-play-state flag on the session that the restage and commit paths read.

src/core/driver/mod.rs
```diff
@@ -413,6 +557,24 @@ impl Driver {
...
+    // pump.
+    #[cfg(any(
+      feature = "local-files",
+      feature = "subsonic",
+      feature = "qobuz",
+      feature = "youtube"
+    ))]
+    if let Some(pause_after) = queue_device_recovered {
+      if pause_after {
+        app.dispatch(IoEvent::PausePlayback);
```
LargeModGames: Two shapes where this settle step misses.

1. Slot mid-download: `advancing` is latched and `native_queue_advance_due` needs `!advancing`, so the advance block above does not fire. This `PausePlayback` pauses the empty reopened sink, then `finish_decoded_fetch` calls `play_file`, which ends in `sink.play()`, and the queue continues on the speakers under "paused here".
2. Queue empty over a suspended Spotify, shuffled Spotify or Radio context: `advance_native_queue` runs `resume_or_finish`, which dispatches `ResumeSpotifyContext` (or the radio equivalent) behind this pause. The pause is a no-op on the idle player and the context resumes at full volume on the new default.

Same fix as the context arm: a desired-play-state flag the resume and commit paths consult, instead of an event that races them.

src/core/driver/mod.rs
```diff
@@ -388,6 +388,150 @@ impl Driver {
...
+        if reopened {
+          // The fresh sink is empty, so the advance block below fires and the
+          // next queued item starts: settle its play/pause state after it, not
+          // here, or the queue plays out loud on a device the user just left.
+          queue_device_recovered = Some(pause_after);
+        } else {
+          // No device to play on. Hand the slot to the same teardown a drained
+          // queue uses: clearing `queue_now` here would strand the suspended
+          // context (latched `advancing`) and the remaining queued items.
+          app.dispatch(IoEvent::FinishNativeQueue);
```
LargeModGames: No latch here. `FinishNativeQueue` is dispatched, but nothing `lost` reads changes, so every tick until the pump drains it repeats the 5 s `reopen()` under the App lock and restamps the message. Then `resume_or_finish` restages the suspended context onto the same `LocalPlayer` when it is `Arc::ptr_eq` to the slot's (a suspended Local context shares it): `live_player()` refuses, `play_file` bails, `fail_index` marks the resume index failed, and a one-track context ends with "no playable tracks left", a device error reported as an unplayable library. Latch the failure so the reopen runs once, and have the teardown either reopen the shared player or surface a device error on the restage instead of poisoning the track.

src/core/driver/mod.rs
```diff
@@ -388,6 +388,150 @@ impl Driver {
...
+          let pause_after = $app
+            .$playback
+            .as_ref()
+            .is_some_and(|s| s.player.device_removed() || s.player.is_paused());
+          // Reopening is what makes the session playable again; if even the new
+          // default device will not open there is nothing to restage onto.
+          let reopened = $app
+            .$playback
+            .as_ref()
+            .is_some_and(|s| s.player.reopen().is_ok());
```
LargeModGames: Keeping the reopen on the tick was my call and 5 s is fine as a backstop. The part that is not fine is that a timeout ends the session with no retry. A Bluetooth output on macOS can take 1 to 5 s to negotiate after AirPods go back in the case, so a timeout here is a device that would have opened a second later, and the flag that would trigger another attempt died with `$playback`. On timeout keep the session, and retry on a later tick with a bounded attempt count.

src/infra/audio/player.rs
```diff
@@ -60,35 +123,161 @@ impl LocalPlayer {
...
+  pub fn device_removed(&self) -> bool {
+    self.sink.lock().unwrap().lost.load(Ordering::Relaxed)
+  }
+
+  /// Whether the sink is still connected to something the user can hear. The
+  /// driver's tick polls this to decide when to [`reopen`](Self::reopen).
+  ///
+  /// Two ways to fail (see module docs): the device was removed and cpal told
+  /// us, or the OS quietly moved its default output elsewhere and nobody did.
+  pub fn device_lost(&self) -> bool {
```
LargeModGames, line 276: This holds the sink mutex across a full `default_host().default_output_device()` plus `description()` round trip, and `Driver::tick` calls it for every live session on every tick, which is 16 ms on the Home screen with the banner gradient on. That is roughly 60 WASAPI enumerator and property-store reads a second (CoreAudio also enumerates every supported input and output config), on the UI thread, under the App lock, while `position()` / `is_paused()` / `is_finished()` on the render path wait on the same mutex. Throttle the name compare to once a second, and release the sink lock before the cpal query.

src/infra/audio/player.rs
```diff
@@ -158,34 +337,40 @@ impl LocalPlayer {
 
   /// Play a prepared stream, replacing whatever was playing. The clear waits
   /// for the audio thread to drop the previous source: call it off the `App`
-  /// lock (see `stop_detached`).
+  /// lock (see `stop_detached`). A no-op once the device is gone.
   #[cfg(any(feature = "internet-radio", feature = "qobuz"))]
   pub fn play_prepared(&self, stream: PreparedStream) {
```
LargeModGames: `play_prepared` did not get the post-clear `Arc::ptr_eq(&sink, &self.player())` re-check that `play_file` got. `bounded(clear)` returns promptly (the identity check is only on the poll), so a stream prepared across a concurrent `reopen()` is appended to the discarded sink and `play()` is called on it. The fresh sink reports `device_lost() == false`, so recovery never fires again: the track shows as playing and the position never advances. For radio that is forever, since radio never polls `is_finished`.

src/infra/queue/mod.rs
```diff
@@ -373,13 +373,59 @@ fn restore_in_place<T>(items: &mut Vec<T>, backup: &ShuffleBackup, current: usiz
 pub async fn replay_file(
   player: std::sync::Arc<crate::infra::audio::LocalPlayer>,
   path: std::path::PathBuf,
+  resume: Option<ResumePoint>,
```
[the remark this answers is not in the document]
LargeModGames: @coderabbitai `replay_file` already carries that gate: the `#[cfg(any(feature = "local-files", feature = "subsonic", feature = "qobuz", feature = "youtube"))]` attribute sits directly above it at `src/infra/queue/mod.rs:367-372`, the same gate `ResumePoint` and `restage` use. The slim and headless legs of this push compiled and passed: Check, Clippy and Test Suite for `slim` and `headless` are green on `e98e87a`. Please verify against the file.

src/runtime/startup.rs
```diff
@@ -339,13 +339,10 @@ pub(super) async fn launch_ui(boot: Boot) -> Result<()> {
...
-  // Gated on whether streaming will be attempted (the player itself now
-  // initializes in the background): registering media keys for a session
-  // whose native init later fails is harmless — the handlers just no-op.
-  // macOS plays no decoded source, so without native streaming the keys would
-  // only be taken away from the other players.
+  // Registered without a Spotify session, like MPRIS and SMTC, so decoded
+  // sources get media keys too; the handlers no-op while nothing plays.
   #[cfg(all(feature = "macos-media", target_os = "macos"))]
-  let macos_media_manager: Option<Arc<macos_media::MacMediaManager>> = if streaming_attempted {
+  let macos_media_manager: Option<Arc<macos_media::MacMediaManager>> =
```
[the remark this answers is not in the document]
LargeModGames: @coderabbitai I asked for this registration in review. It matches the Windows SMTC path, which registers at startup regardless of playback, and the macOS behavior with a Spotify session, where the handlers were registered before anything played too. macOS routes remote commands to the app whose Now Playing center reports a playing state; `MacMediaManager` sets that state from the playback snapshot and reports `Stopped` with no info when nothing plays, so an idle spotatui is not the Now Playing app. Registering and removing the handlers dynamically is a separate change, outside this PR. Please verify the state handling in `src/infra/macos_media.rs` (`SetStopped`, `setPlaybackState`).

src/infra/queue/dispatch.rs
```diff
@@ -894,7 +952,13 @@ async fn resume_or_finish(app: &Arc<Mutex<App>>) {
       if let Some(player) = queue_player {
         player.stop();
       }
-      if let Some(uri) = station.uri.clone() {
+      // A station cannot resume paused: after a device removal it stays off.
+      if !playing {
+        app
+          .lock()
+          .await
+          .set_status_message("Audio device changed - radio not resumed.".to_string(), 8);
```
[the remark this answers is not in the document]
LargeModGames: @coderabbitai Changed in `9b6cef4` to the wording you proposed: `Playback is paused - radio not resumed.` Please verify against `src/infra/queue/dispatch.rs`.

src/infra/queue/dispatch.rs
```diff
@@ -629,16 +644,28 @@ async fn finish_decoded_fetch(
...
     .await
     .map(|r| r.map_err(|e| e.to_string()))
     .unwrap_or_else(|e| Err(e.to_string()));
-  if let Err(e) = played {
+  let mut guard = app.lock().await;
+  if !matches!(
+    guard.queue_now.as_ref(),
+    Some(QueueNowPlaying::Decoded(d)) if d.fetch_id == fetch_id
+  ) {
+    return;
```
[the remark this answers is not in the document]
LargeModGames: @coderabbitai The stage already guards that window: `LocalPlayer::stage_file` re-reads the sink after the decode and bails with "audio output device changed while decoding" when `recover_device` swapped it, so nothing is appended to a discarded sink there. What is left is the gap between the append and the finalize under the lock. A reopen in that gap drops the old sink with its audio thread, so no audio plays from it, and `resume()` on the fresh empty sink makes the next tick read end-of-track and advance to the next queued item. That is how this PR treats a queue-slot device change on purpose: the interrupted item is skipped, not restaged (see the queue block in `Driver::tick`), because there is no replay event for a queued item. A second identity check in `finish_decoded_fetch` would only move that gap a few microseconds later. Please verify `stage_file` and the driver queue block.

LargeModGames:
@coderabbitai review

src/infra/queue/dispatch.rs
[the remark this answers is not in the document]
LargeModGames: @coderabbitai `e9b4f61` adds the guard from your review of `cf2b3fb` too: `LocalPlayer` carries a stage generation, `stage_file` and `stage_prepared` bump it at entry and bail before the append when a later stage started on the same player. Together with the sink identity check before the append, a stage that was overtaken by a reopen or by a later track never lands. Please verify.

src/infra/audio/player.rs
```diff
@@ -98,24 +439,18 @@ impl LocalPlayer {
...
-  }
+    // The decode is long enough for the driver's tick to have reopened the
+    // device under us, or for a later stage to have cleared the sink for its
+    // own track; appending now would play into nothing, or play stale audio.
+    if !Arc::ptr_eq(&sink, &self.player()) {
+      anyhow::bail!("audio output device changed while decoding");
+    }
+    if self.stage_superseded(stage) {
+      anyhow::bail!("superseded by a later track");
+    }
```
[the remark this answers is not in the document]
LargeModGames: @coderabbitai Done in `0106f36`. The player now holds a `stage_transition` lock across each of a stage's two sink transitions, with the generation checked in front of each: `clear_for_stage` refuses to clear once a later stage began, and `append_for_stage` refuses to append once a later stage began or the device was reopened. The decode runs between the two, outside the lock. Please verify `src/infra/audio/player.rs`.

PR #499, opened 2026-08-28
fix(local): scan nested folders so a two-level music library is browsable

alessandro-zanni:
## The bug

With `behavior.local_music_path: ~/Music` and a library organised as

```
~/Music/Jamendo/Playlist - Fresh & New/001-….mp3   (73 files)
```

the local-files library listed `Jamendo` with a **0** track count, and selecting it showed an **empty track table**.

`LocalSource::playlists` walked only the immediate subdirectories of the root and counted only the audio sitting *directly* inside them (`count_audio_files`, non-recursive). `Jamendo` holds no mp3s of its own, so it got `track_count: 0` — and `tracks()`, reading that same flat directory, found nothing. Any `Artist/Album/…` layout hit this too.

## The fix

`playlists()` now walks the tree and returns one playlist per directory that **actually holds audio**, named by its path relative to the music root:

```
Jamendo/Playlist - Fresh & New   73
```

Directories that only contain other directories are no longer listed — there is nothing to play in them, and listing them was the whole confusion.

`tracks()`, the URIs, and playback are **unchanged**: the playlist URI already pointed at whatever folder the entry named, it now names the folder that holds the files.

Guards in the walk:

- `DirEntry::file_type` instead of `Path::is_dir` — it does not dereference, so a **symlinked directory is not descended into** and a link loop cannot hang the scan.
- A subdirectory that cannot be read (permissions) is **skipped**, not fatal — one denied folder must not hide the rest of the library. An unreadable *root* stays fatal, as before.
- Hidden entries (`.Trash`, `.DS_Store`) skipped; `MAX_SCAN_DEPTH = 8` caps the descent.

The synthetic `(root)` entry keeps its **non-recursive** count on purpose, so it stays consistent with what `tracks()` returns for that same URI.

## Tests

Three added (`playlists_finds_nested_folders_with_audio`, `playlists_omits_folders_with_no_direct_audio`, `playlists_stops_at_max_scan_depth`); `tools/gates.count` baseline moved 1753 → 1756 accordingly.

`playlists_returns_one_entry_per_subdir` needed an audio file added to each of its two folders — it previously asserted that *empty* folders are listed, which is exactly the behaviour this PR changes.

## Verified

`cargo fmt`, slim clippy, and the full slim suite (907 passed) are green. Also scanned the real `~/Music` with a throwaway `#[ignore]` test (removed before commit): `Jamendo/Playlist - Fresh & New` → 73 tracks, `tracks()` → 73 rows.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PLw5FNaW6UNgFmM2dr1NC1

## Summary by CodeRabbit

- **Bug Fixes**
  - Local music scanning now discovers audio in nested folders, up to a safe depth limit.
  - Playlists are created only for folders containing playable audio, using their relative paths as names.
  - Empty container folders are omitted.
  - Linked folders are scanned once, while hidden entries are excluded consistently.
  - Unreadable folders and entries are skipped without interrupting the scan.
  - Playlist track counts now match the tracks displayed.


LargeModGames:
Thanks for this. The fix is correct for the reported layout and the three new tests cover it. I read the branch and found a few regressions the recursive walk introduces, plus some smaller items. Details below, ordered by weight.

## Blocking

**1. Symlinked directories and junctions vanish from the listing** (`src/infra/local/mod.rs:199`, `:571`)

Before the PR `path.is_dir()` dereferenced, so a symlinked directory (or a Windows junction, `mklink /J`) was listed and counted. Now `file_type().is_dir()` and `is_file()` are both false for the link, so the entry matches neither arm and the whole linked tree disappears with no error. A NAS mounted under `~/Music/NAS` is a common layout. The doc and the changelog say links are "not followed", which reads as "still listed, not descended". Either keep them in the listing (and guard the descent some other way) or state the drop plainly in both places.

**2. `track_count` disagrees with `tracks()` for the same URI** (`src/infra/local/mod.rs:181`, `:275`, `:581`)

`playlists()` now counts with the non-dereferencing `file_type().is_file()` plus the hidden filter. `tracks()` is unchanged: it uses the dereferencing `path.is_file()` and has no hidden filter. So a folder with 10 real files and 5 symlinked tracks shows "(10 tracks)" in the sidebar and 15 rows in the table. An exFAT drive with `._01 Track.mp3` AppleDouble siblings counts 12 and shows 24 rows. A folder whose tracks are all symlinks has `audio_count == 0` and is never listed. The new doc line "Its count stays non-recursive so it matches what `tracks` returns" is false today. One entry policy in both functions, plus a regression test for a hidden file and a symlinked file, closes this. CodeRabbit flagged the same thing.

**3. Empty results now blame the config path** (`src/infra/local/mod.rs:561`, `:581`, `:584`)

Three cases now return an empty list where the old code returned folders with `0`:

- a library of only unsupported formats (`.opus` for example);
- an unreadable subtree (`let Ok(entries) = read_dir(dir) else { return }` and `entries.flatten()` drop every error, at the root level too, where a per-entry error was fatal before);
- a library deeper than `MAX_SCAN_DEPTH`.

In all three the sidebar renders "Local Files (no folders - set behavior.local_music_path)" even though the path is correct, and nothing tells the user why. At minimum, a `log::warn` per skipped directory and per depth-cap truncation. The old "folder with 0" was at least a signal that the path was seen.

## Smaller

**4. Stale docs.** The `LocalSource` struct doc (`src/infra/local/mod.rs:132`), `src/infra/local/dispatch.rs:533`, and `src/core/app/mod.rs:301` still describe one playlist per immediate subdirectory. Only the `playlists()` method doc was updated. The struct doc is the one `cargo doc` shows first. Please also name the 8-level limit in the `playlists()` contract (CodeRabbit asked for the same in the changelog).

**5. `MAX_SCAN_DEPTH` comment** (`src/infra/local/mod.rs:529`). The `///` line, an empty `///`, then a `//` block that starts with "ponytail". The word appears nowhere else in the repo and reads as an edit artifact. One `///` line with the reason (depth cap instead of a visited-inode set) is enough.

**6. The root loop duplicates `collect_audio_dirs`** (`src/infra/local/mod.rs:195`). It is the same body at depth 0. A later rule added to one silently misses the other. Calling `collect_audio_dirs(&self.root, 0, &mut dirs)` and taking the root entry out as the `(root)` sentinel removes the copy, at the cost of one extra `read_dir` on the root for the error message. Judgement call, not blocking.

**7. Sidebar clipping.** The `/`-joined name goes into `PlaylistInfo.name` and is drawn with no ellipsis, so `Artist/Album One` and `Artist/Album Two` clip to the same prefix in a narrow sidebar and the "(N tracks)" suffix is the first thing lost. Cosmetic. Also note that Lua and MCP consumers now get a path-like `name` with a hardcoded `/` on every platform.

**8. Pump cost.** The walk is now one `read_dir` per directory in the tree, still awaited inline on the serial IoEvent pump and re-run on every switch into the Local source. The blocking I/O is pre-existing (the TODO in the doc names it), so not blocking here, but worth a note that a large network-mounted library pays the full walk each time.

## Housekeeping

- Every other entry under `### Fixed` links an issue. If you open one for the nested-library bug, link it from the changelog line.

LargeModGames:
@all-contributors please add @alessandro-zanni for code

Issue #494, opened 2026-08-27
Name the decoded-source cfg predicates with alias features


LargeModGames:
**Is your feature request related to a problem? Please describe.**
About 80 `#[cfg(any(feature = "local-files", feature = "subsonic", feature = "qobuz", feature = "youtube"))]` attributes across some 20 files spell out "a queueable decoded source is compiled in" by hand, and the five-way variant with `internet-radio` is exactly `feature = "audio-decode"`. Adding Qobuz meant editing every one of them (#489), and a missed list compiles fine but silently drops that source from one feature combination that no CI leg exercises.

**Describe the solution you'd like**
One implied feature in `Cargo.toml`, for example `decoded-queue = []`, listed inside `local-files`, `subsonic`, `qobuz`, and `youtube`. The four-way lists become `#[cfg(feature = "decoded-queue")]` and the five-way lists become `#[cfg(feature = "audio-decode")]`. A sixth source then touches `Cargo.toml` and nothing else.

**Describe alternatives you've considered**
Leaving the lists as they are; the next source repeats the same sweep.

**Additional context**
Mechanical, no behaviour change; from the review of #489.

topongo:
i'm looking into it. will CI pipelines check for uncommon feature intersections? if not i will run them locally: changing so many feature gating directives is going to be risky

LargeModGames:
> i'm looking into it. will CI pipelines check for uncommon feature intersections? if not i will run them locally: changing so many feature gating directives is going to be risky

No, CI only compiles the decoded sources all at once (all-sources and macos), so please run cargo clippy and cargo test locally with each single source, with and without streaming, before you open the PR.

PR #523, opened 2026-09-11
refactor: feature aliasing

topongo:
# Summary
this aims to solve #494, by adding some "meta-features". meta-features indicates empty cargo features, referenced by other "real" ones. when inserting `#[cfg(feature = "meta-feature1")]` is the equivalent of `#[cfg(any(feature = "f1", feature = "f2", ...))]`. this simplifies complex feature gates.

the added features are the following:
- `audio-decode-queue`: `qobuz`, `youtube`, `local-files`, `subsonic`
  audio features that needs to be decoded using `rodio`, but also supports a non-infinite (unlike internet-radio) queue.
- `queue`: `audio-decode-queue`, `streaming` 
  features supporting a non-infinite queue.
- `queue-download`: `subsonic`, `youtube`, `qobuz`
  features that needs full file download before playback that also supports non-infinite queue.
- ~~`onboarding`: `local-files`, `subsonic`, `youtube`
  features that requires onboarding.~~

# Testing
i did run `cargo {clippy,test} --no-default-features` on the following feature sets: empty (core only), each source on its own (`streaming`, `qobuz`, `youtube`, `local-files`, `subsonic`), each source with `streaming`, each source with `streaming` and `tui`.
everything passed.

edit: removed `onboarding` feature, as it was used only once and it was confusing.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated playback, audio decoding, queue, and download feature configuration for more consistent builds across supported sources.
  * Updated queue and playback components to use unified capability settings while preserving runtime playback behavior.
  * Improved feature-specific handling for queue downloads, playback quality, and suspended-context previews.

* **Tests**
  * Qobuz playback tests now report failures when playback cannot be started.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,7 +158,7 @@ tui = ["dep:ratatui", "dep:crossterm", "dep:tui-bar-graph", "dep:colorgrad"]
 # plumbing); the frontend itself has not landed yet.
 gui = []
 self-update = ["dep:self_update", "dep:sha2", "dep:hex"]
-streaming = ["librespot-core", "librespot-playback", "librespot-connect", "librespot-oauth", "librespot-metadata", "librespot-protocol", "protobuf"]
+streaming = ["librespot-core", "librespot-playback", "librespot-connect", "librespot-oauth", "librespot-metadata", "librespot-protocol", "protobuf", "queue"]
 # Audio backend features
 alsa-backend = ["streaming", "librespot-playback/alsa-backend"]
 pulseaudio-backend = ["streaming", "librespot-playback/pulseaudio-backend"]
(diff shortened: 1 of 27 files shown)


LargeModGames:
Codecov can be ignored

topongo:
> Codecov can be ignored

thanks, but the first coderabbit submission is right. i'm fixing.

LargeModGames (changes requested):
One change before merge: replace the `#[cfg(feature = "tui")]` on `DecodedQueuePlayback::quality` with `#[allow(dead_code)]` (like `tempfile` next to it) and drop the two matching `tui` gates in `dispatch.rs`. The queue engine must not depend on the terminal frontend.

Optional: `src/core/onboarding.rs:71` can use `not(feature = "queue-download")`, and `audio-decode-queue = ["audio-decode"]` plus `queue-download = ["audio-decode-queue", "dep:tempfile"]` would close the two CodeRabbit threads.

topongo:
> Optional: `src/core/onboarding.rs:71` can use `not(feature = "queue-download")`, and `audio-decode-queue = ["audio-decode"]` plus `queue-download = ["audio-decode-queue", "dep:tempfile"]` would close the two CodeRabbit threads.

ok, this make sense. i actually tried to do this, but then i scrapped the idea.

> One change before merge: replace the `#[cfg(feature = "tui")]` on `DecodedQueuePlayback::quality` with `#[allow(dead_code)]` (like `tempfile` next to it) and drop the two matching `tui` gates in `dispatch.rs`. The queue engine must not depend on the terminal frontend.

i noticed that `quality` is only used by the tui, that's why i gated it.
what about `cfg_attr(not(feature = "tui"), allow(dead_code))`?

topongo:
another thing, i was thinking about prefixing meta-features with underscores, so that users know not to call them while building. or in any case to let users know that enabling those features doesn't actually add any concrete feature to the binary.
what do you think about it?

LargeModGames:
> i noticed that `quality` is only used by the tui, that's why i gated it.
> what about `cfg_attr(not(feature = "tui"), allow(dead_code))`?

Yes, `#[cfg_attr(not(feature = "tui"), allow(dead_code))]` is the form this repo already uses (`src/core/mod.rs`, `src/infra/mod.rs`), and it is better than the plain `allow`: every tui leg still lints the field in full. Use it on the field and drop the two `tui` gates in `dispatch.rs`. With the field in every build, the write and the `let (tmp, quality)` binding can be unconditional.

LargeModGames:
> another thing, i was thinking about prefixing meta-features with underscores, so that users know not to call them while building. or in any case to let users know that enabling those features doesn't actually add any concrete feature to the binary.
> what do you think about it?

No prefix. The repo already has three implementation features without one: `audio-decode`, `art-decode` and `dj-core`. The comment above each one is what tells a user not to enable it by hand, and three prefixed features next to those would be inconsistent. With the prerequisites you are adding, enabling `audio-decode-queue` or `queue-download` alone also compiles and just adds no source. Keep the comments, and say in them that the sources enable these features and nothing else should.

topongo:
features

for future reference

LargeModGames:
> for future reference

The diagram shows the missing edge: `audio-decode-queue = audio-decode ∩ queue`, but Cargo.toml only lists `audio-decode`. `--features telemetry,queue-download` fails with 27 errors on `queue`-gated items. `audio-decode-queue = ["audio-decode", "queue"]` closes it and the open CodeRabbit thread at `dispatch.rs:526`.

Take CodeRabbit's comment fix. Skip its test: `finish_decoded_fetch` needs a real output device, and the `quality` write is unchanged from main.

LargeModGames (approved):
`2a5edf0` fixes it: `queue-download` and `audio-decode-queue` now compile alone. `queue` alone still fails on the empty match in `queue_now_track` (`src/core/app/queue.rs:171`), but a `queue`-only build is out of contract, so I am fine either way.

PR #524, opened 2026-09-12
feat: cancel superseded native-queue downloads

### LINKED ISSUE #491: Cancel superseded native-queue downloads
`play_queued_subsonic`, `play_queued_qobuz`, and `play_queued_youtube` in `src/infra/queue/dispatch.rs` detach their download task and keep no handle. A skip republishes the queue slot, but the old download runs to completion; the stale result is discarded by the `fetch_id` check only when it lands. Fast skips through hi-res Qobuz tracks keep several downloads and tempfile writes alive at the same time.

An abort handle on `DecodedQueuePlayback`, stored under the same lock that stamps `fetch_id`, and aborted whenever the slot is republished or cleared. One change covers all three lanes. Add a regression test for a skip during a pending download.

**Describe alternatives you've considered**
Keeping the current behaviour: correct, but it wastes bandwidth and disk on every fast skip.

**Additional context**
Raised by CodeRabbit on #489; I deferred it there because the fix belongs to the shared queue engine, not to one source.

RohithPariki:
## Summary
Fixes #491.
This adds robust background task cancellation for native-queue downloads (Subsonic, Qobuz, YouTube) when the user skips tracks rapidly. It ensures we don't leak tokio tasks or needlessly waste network bandwidth/disk I/O on downloads that will be discarded.

## Problem
In play_queued_subsonic, play_queued_qobuz, and play_queued_youtube, the tokio::spawn download task runs completely detached. If a user skips quickly, the queue slot is republished but the old task runs to completion. The stale result is only dropped when the download finally finishes and finish_decoded_fetch sees the fetch_id mismatch.

## Solution
I implemented a Drop-driven cancellation pattern:
1. Created DownloadAbortHandle, a thin wrapper around tokio::task::AbortHandle that calls .abort() on Drop.
2. Added abort_handle: Option<DownloadAbortHandle> to DecodedQueuePlayback.
3. In the dispatch handlers, after spawning the background task, we take the App lock one more time. If the slot still carries the matching fetch_id, we inject the abort handle. If it's already advanced, we .abort() it immediately.
4. Because queue_now is overwritten on skips (or cleared on teardown), the old DecodedQueuePlayback is naturally dropped, bringing our DownloadAbortHandle with it and cleanly cancelling the background task instantaneously.

## Testing
Added the regression test test_queue_skip_aborts_pending_download to src/infra/queue/dispatch.rs that explicitly verifies a skip clears the slot and correctly aborts the pending download task.

## Related Issue
Fixes #491

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cancellation of pending downloads for Subsonic, Qobuz, and YouTube playback.
  * Downloads now stop promptly when their queue slot is cleared or replaced.
  * Prevented background download tasks from continuing after playback items are removed or updated.
  * Improved queue cleanup during playback changes, reducing unnecessary downloads and resource usage.

### DIFF
--- a/src/infra/queue/dispatch.rs
+++ b/src/infra/queue/dispatch.rs
@@ -766,6 +766,8 @@ async fn publish_decoded(
     fetch_id: next_fetch_id(),
     #[cfg(any(feature = "subsonic", feature = "qobuz", feature = "youtube"))]
     tempfile,
+    #[cfg(any(feature = "subsonic", feature = "qobuz", feature = "youtube"))]
+    abort_handle: None,
     quality: None,
   }));
   guard.set_status_message(format!("\u{266a} {name} (queue)"), 4);
--- a/src/infra/queue/dispatch.rs
+++ b/src/infra/queue/dispatch.rs
@@ -377,15 +377,30 @@ async fn play_queued_subsonic(app: &Arc<Mutex<App>>, track: &TrackInfo, uri: &st
   let fetch_id = publish_pending_decoded(app, &player, track).await;
   // Fetch off the IoEvent pump: awaiting the download here would freeze every
   // other event (skips included, for every source) for its whole duration.
-  let app = Arc::clone(app);
+  let app_clone = Arc::clone(app);
   let uri = uri.to_string();
   let name = track.name.clone();
-  tokio::spawn(async move {
+  let handle = tokio::spawn(async move {
     let result = crate::infra::subsonic::dispatch::download_for_queue(&source, &uri)
       .await
       .map(|tmp| (tmp, None));
-    finish_decoded_fetch(&app, fetch_id, result, &name).await;
+    finish_decoded_fetch(&app_clone, fetch_id, result, &name).await;
   });
+  
+  let abort_handle = handle.abort_handle();
+  {
+    let mut guard = app.lock().await;
+    let mut injected = false;
+    if let Some(crate::infra::queue::QueueNowPlaying::Decoded(ref mut d)) = guard.queue_now {
+      if d.fetch_id == fetch_id {
+        d.abort_handle = Some(crate::infra::queue::DownloadAbortHandle(abort_handle.clone()));
+        injected = true;
+      }
+    }
+    if !injected {
+      abort_handle.abort();
+    }
+  }
   true
 }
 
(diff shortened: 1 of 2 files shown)


LargeModGames (changes requested):
Thanks for the PR, the design is the right one. A `Drop` on the slot field covers every site that replaces or clears `queue_now`, the second lock after the spawn closes the race with a fast skip, and `finish_decoded_fetch` updates the slot in place so the handle survives a completed download. Four things block CI, and two are cleanup.

**CI failures**

1. **Rustfmt.** The blank line after each `tokio::spawn` block carries two spaces, and one new line is over the width limit. `cargo fmt --all` corrects both.
2. **Gates ratchet.** The new `#[tokio::test]` moves `test_attribute_total` from 1862 to 1863. Bump the value in `tools/gates.count` in this PR; the ratchet only lets it rise, so this is expected.
3. **Clippy (`empty_line_after_doc_comments`).** `DownloadAbortHandle` landed between the doc comment of `DecodedQueuePlayback` and its `#[cfg]`, so the doc comment now documents the wrong struct. Move the new struct and its `Drop` impl above that doc comment and give it a one-line doc of its own. Only the macOS leg and `Clippy (all-sources)` catch this, because the other legs do not enable subsonic, qobuz, or youtube.
4. **The test opens an audio device.** `LocalPlayer::new()` opens the default output, and CI runners have none, so `Test Suite (all-sources)` and `Coverage` panic with `opening default audio output device`. The test also stores the handle by hand, so it only proves that `Drop` calls `abort`. A test of `DownloadAbortHandle`'s `Drop` alone, next to the struct in `mod.rs`, proves the same thing with no player. Please also drop the `test_` prefix; test names in this repo are behavior sentences, for example `dropping_the_abort_handle_cancels_the_download_task`.

**Cleanup**

- The three identical blocks after the spawn can be one `async fn attach_abort_handle(app, fetch_id, handle)`. Inside it, the `injected` flag and the `abort_handle.clone()` are not needed: store the handle when the `fetch_id` matches, else abort it.
- The `app` to `app_clone` rename is churn. Keep `app` and give only the copy moved into the closure a new name.

Once these land I will run the full build locally and merge.

